PR #6, opened 2019-08-27
Recognize wp_die as earlyTerminatingFunction

szepeviktor:
First we need https://github.com/phpstan/phpstan/pull/2392


szepeviktor:
@canvural Thank you for https://github.com/phpstan/phpstan-src/commit/3a1e0a9b9e5e00c9bcddec6f464f7d44a575388c

PR #10, opened 2019-12-06
Fix for FixWpStubs polyfill method

mallardduck:
Current Results:

```
Generating autoload files
> PHPStan\WordPress\Composer\FixWpStubs::php73Polyfill
Removing duplicate is_countable() ...
Script PHPStan\WordPress\Composer\FixWpStubs::php73Polyfill handling the post-update-cmd event terminated with an exception


  [ErrorException]
  file_get_contents(/Users/danpock/sites/wptest/vendor/giacocorsiglia/wordpress-stubs/wordpress-stubs.php): failed to open stream: No such file or
   directory
```

This file would be expected if the repo were using `giacocorsiglia`, however it's currently using from `php-stubs`.

### DIFF
--- a/src/Composer/FixWpStubs.php
+++ b/src/Composer/FixWpStubs.php
@@ -9,7 +9,7 @@ use Composer\Script\Event;
 
 class FixWpStubs
 {
-	const STUBSFILE = '/giacocorsiglia/wordpress-stubs/wordpress-stubs.php';
+	const STUBSFILE = '/php-stubs/wordpress-stubs/wordpress-stubs.php';
 
 	public static function php73Polyfill(Event $event): int
 	{


szepeviktor:
Hello @mallardduck !

Thank you for the PR.

The thing is I've released php-stubs because Giacocorsiglia's one is not updated and contains compat code. Do if you use stubs from php-stubs **you do not need** the polyfill at all.

mallardduck:
Ahhh! Great to know @szepeviktor, so - instead of this PR - I can simply exclude the polyfills from my composer.json update/install scripts?

szepeviktor:
Oh YESSS!

mallardduck:
Rock on! That's what I'll do then, was just going off the example folder. 😄

PR #17, opened 2020-06-08
Fix the deprecation notice

ondrejmirtes:
This is how you get rid of the error. Please release this. Thanks.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -24,7 +24,7 @@ services:
         tags:
             - phpstan.broker.dynamicFunctionReturnTypeExtension
 parameters:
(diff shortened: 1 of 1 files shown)


szepeviktor:
@ondrejmirtes Thank you.

Does it mean that this package won't be compatible with 0.12.0 - 0.12.25?

ondrejmirtes:
Oh yes, I should have bumped the version in composer.json as well. I’ll do it once I’m in front of the computer again. You can tag it as 0.6.1.

szepeviktor:
Doing it right now :)

Thank you.

PR #18, opened 2020-08-09
Added some core constants.

fullbl:
I added some constants to avoid errors from phpstan!


szepeviktor:
First thank you for your contribution!

As you can see in the README

> Avoid using core constants, use core functions

Please consider avoiding core functions, they are not meant to be used in plugins and themes.
Although there is some controversy there.

We have _naughty_ project for constants: https://github.com/php-stubs/wordpress-globals
Please consider **not** using it 👍

PR #21, opened 2020-09-04
Remove constants from bootstrap.php

dingo-d:
This PR removes hardcoded constants from `bootstrap.php` file (removes the entire file), and instead relies on `php-stubs/wordpress-globals`.

I did get some errors when running `composer test:cs`

```php
FILE: ...wordpress/src/WpThemeMagicPropertiesClassReflectionExtension.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 18 | ERROR | [x] Property
    |       |     \PHPStan\WordPress\WpThemeMagicPropertiesClassReflectionExtension::$properties
    |       |     does not have native type hint for its value but it
    |       |     should be possible to add it based on @var
    |       |     annotation "array<int, string>".
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------


FILE: ...pstan-wordpress/src/IsWpErrorFunctionTypeSpecifyingExtension.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 22 | ERROR | [x] Property
    |       |     \PHPStan\WordPress\IsWpErrorFunctionTypeSpecifyingExtension::$typeSpecifier
    |       |     does not have native type hint for its value but it
    |       |     should be possible to add it based on @var
    |       |     annotation "\PHPStan\Analyser\TypeSpecifier".
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

Should these be fixed here or in another PR?


dingo-d:
Ok, from our conversation, it should be the other way around (removing duplicates in the other package).

szepeviktor:
These constant are in `bootstrap.php` because they are *meant to be used* in plugins and themes.

PR #23, opened 2020-09-20
Add WP-CLI early terminating calls

### LINKED ISSUE #22: Early return not recognized in the case of `\WP_CLI::error`
Hi!

I have a method

```php
public function getComposer(array $args = []): array
{
	if (!isset($args['config_path'])) {
		if (function_exists('\add_action')) {
			$composerPath = $this->getProjectRootPath() . '/composer.json';
		} else {
			$composerPath = $this->getProjectRootPath(true) . '/composer.json';
		}
	} else {
		$composerPath = $args['config_path'];
	}

	$composerFile = file_get_contents($composerPath);

	if ($composerFile === false) {
		\WP_CLI::error("The composer on {$composerPath} path seems to be missing.");
	}

	return json_decode($composerFile, true);
}
```

And I'm getting an error

```bash
Parameter #1 $json of function json_decode expects string,
string|false given.
```

But above the `json_decode` there is a check

```php
if ($composerFile === false) {
	\WP_CLI::error("The composer on {$composerPath} path seems to be missing.");
}
```

And `\WP_CLI::error` will `exit()` out, and terminate script execution.

dingo-d:
The ones that have `exit()` calls in them.

Fixes #22 

I'm not sure if the functions from the [utils.php](https://github.com/wp-cli/wp-cli/blob/master/php/utils.php) be included as well. Those are utility functions, not sure if they are used in userland code, or just for internal WP-CLI purposes.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -43,5 +43,3 @@ parameters:
         \WP_CLI:
             - WP_CLI::error
             - WP_CLI::halt
-            - WP_CLI::launch
-            - WP_CLI::runcommand
--- a/extension.neon
+++ b/extension.neon
@@ -39,3 +39,9 @@ parameters:
         - wp_send_json_error
         - wp_nonce_ays
         - dead_db
+    earlyTerminatingMethodCalls:
+        \WP_CLI:
+            - WP_CLI::error
+            - WP_CLI::halt
+            - WP_CLI::launch
+            - WP_CLI::runcommand


extension.neon
```diff
@@ -39,3 +39,9 @@ parameters:
         - wp_send_json_error
         - wp_nonce_ays
         - dead_db
+    earlyTerminatingMethodCalls:
+        \WP_CLI:
+            - WP_CLI::error
+            - WP_CLI::halt
+            - WP_CLI::launch
+            - WP_CLI::runcommand
```
szepeviktor, line 47: I think these commands terminate execution only with a complex condition.

https://github.com/wp-cli/wp-cli/blob/d7df11b43e9ece6990ac42371aadac2cbd9d6554/php/class-wp-cli.php#L1270-L1271

`earlyTerminatingMethodCalls` is for methods unconditionally terminating execution.

Even if you write a PHPStan service for `WP_CLI::runcommand` you cannot figure out what the command's exit status will be.
:) So never enable `exit_error` to keep your code testable.

szepeviktor (approved):

szepeviktor (changes requested):
Please remove those two.

extension.neon
szepeviktor: Added this warning to the README

dingo-d:
@szepeviktor Thanks for the clarification, wasn't sure if any command should be added. I've removed those two 👍

szepeviktor:
Thank you.

szepeviktor:
https://github.com/szepeviktor/phpstan-wordpress/releases/tag/v0.6.4 🎉

PR #27, opened 2020-11-03
Add support to `\wpdb::__get` magic method

andreasciamanna:
PHPStan would show an error when accessing properties such as `$wpdb->dbname` as they are protected properties.

However, `\wpdb` implements `__get`, which means these properties are accessible.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -4,6 +4,10 @@ services:
         tags:
             - phpstan.typeSpecifier.functionTypeSpecifyingExtension
     -
+        class: PHPStan\WordPress\wpdbMagicPropertiesClassReflectionExtension
+        tags:
+            - phpstan.broker.propertiesClassReflectionExtension
+    -
         class: PHPStan\WordPress\WpThemeMagicPropertiesClassReflectionExtension
         tags:
             - phpstan.broker.propertiesClassReflectionExtension
--- a/src/wpdbMagicPropertiesClassReflectionExtension.php
+++ b/src/wpdbMagicPropertiesClassReflectionExtension.php
@@ -1,0 +1,49 @@
+<?php
+
+/**
+ * Accept magic properties of wpdb.
+ */
+
+declare(strict_types=1);
+
+namespace PHPStan\WordPress;
+
+use PHPStan\Reflection\ClassReflection;
+use PHPStan\Reflection\PropertyReflection;
+use PHPStan\Reflection\Dummy\DummyPropertyReflection;
+
+class wpdbMagicPropertiesClassReflectionExtension implements \PHPStan\Reflection\PropertiesClassReflectionExtension
+{
+	/** @var array<int, string> */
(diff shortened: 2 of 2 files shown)


andreasciamanna:
I see the build fails because `wpdbMagicPropertiesClassReflectionExtension` is not in pascal case.

However, the original `wpdb` class has the wrong casing from the start. 

Not sure if I should stick to the correct name or to the expected convention.

szepeviktor:
> I see the build fails because `wpdbMagicPropertiesClassReflectionExtension` is not in pascal case.

Hello Adrea! Let's not care about letters :)

Is these `wpdb` properties meant to be used publicly?
This PHPStan extension should only support public WordPress API-s.
What do you think?

andreasciamanna:
I can't speak for the rest of the worlds, but, in WPML, we have a few instances where we need to access the table schema (e.g. to check and/or fix collation).

That requires writing code such as this:

```php
global $wpdb;

$wpdb->get_var( "
SELECT COLLATION_NAME FROM INFORMATION_SCHEMA.COLUMNS
WHERE TABLE_SCHEMA = '{$wpdb->dbname}'
	AND TABLE_NAME = '{$wpdb->prefix}{$table_name}'
	AND COLUMN_NAME = '{$column_name}'
" );
```

`\wpdb::dbname` is a protected property and PHPStan complains about that, as expected.

The alternative would be to replace `$wpdb->dbname` with `DB_NAME`.  
However, since `\wpdb::dbname` is accessible via `\wpdb::__get`, that makes it, de facto, a public property.

It's my opinion that, if the WP-Core didn't want `\wpdb::dbname` to be accessible, they would have written `__get` so to throw an exception when the requested property is `===dbname`.

szepeviktor:
All right.

Please me someone else who also uses `$wpdb->dbname`.
Until then please add that service to your `phpstan.neon`˛

BTW Your class is very nicely written 😻

andreasciamanna:
I did try adding the class to the services, but I'm clearly missing something (and PHPStan documentation doesn't help with it comes to extensions).

This is now in my `phpstan.neon` file:

```yaml
services:
  -
    class: PHPStan\WordPress\wpdbMagicPropertiesClassReflectionExtension
    tags:
      - phpstan.broker.propertiesClassReflectionExtension
```

As the class is not autoloaded, I've added the file to the `bootstrapFiles` list.

PHPStan sill shows the error about the protected property.

> BTW Your class is very nicely written 

Most likely, because it's 80% a copy of `WpThemeMagicPropertiesClassReflectionExtension` :P

szepeviktor:
Noooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo!

szepeviktor:
That class should go to your own `/tests` folder.

`\WPML\Andrea\Tests\PHPStan\WordPress\wpdbMagicPropertiesClassReflectionExtension` the namespace you already have.

andreasciamanna:
Damn. My fault!

It works, sorry: I was updating a `phpstan.neon` of a different project :)

PR #31, opened 2021-01-03
Define more functions that unconditionally terminate

johnbillion:
These functions and methods all unconditionally terminate.


szepeviktor:
Thank you.

Indeed they do.

PR #32, opened 2021-01-08
Another method that unconditionally terminates

johnbillion:


szepeviktor:
Really https://developer.wordpress.org/reference/classes/wp_ajax_response/send/

PR #34, opened 2021-02-15
Set return type of wp_slash() and wp_unslash().

Chouby:
As `esc_sql()`, `wp_slash()` and `wp_unslash()` returns  `string|array`, but always the same type of the passed parameter.


szepeviktor:
Thank you @Chouby!

Could you rename the class to something like `StringOrArrayDynamicFunctionReturnTypeExtension` ?

Chouby:
It's indeed better. Done!

szepeviktor:
Thank you.

PR #36, opened 2021-03-01
Add dynamic return type extension for shortcode_atts

monojp:
See https://developer.wordpress.org/reference/functions/shortcode_atts/

I had a case where I strictly specified a list of supported shortcode args and used `shortcode_atts` to filter what is "coming in" which PHPStan did not understand. Something like the following but more dynamic in a function
```php

/** @var array{foo: string, baz: ?string} */
$supportedArgs = [
    'foo' => 'bar',
    'baz' => null,
];
$args = shortcode_atts($supportedArgs, $userArgs);
// $args is still array{foo: string, baz: ?string} here independent of what keys are existing in $userArgs

```

### DIFF
--- a/src/ShortcodeAttsDynamicFunctionReturnTypeExtension.php
+++ b/src/ShortcodeAttsDynamicFunctionReturnTypeExtension.php
@@ -1,0 +1,37 @@
+<?php
+
+/**
+ * Set return type of shortcode_atts().
+ */
+
+declare(strict_types=1);
+
+namespace PHPStan\WordPress;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\Type;
+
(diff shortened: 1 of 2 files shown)


src/ShortcodeAttsDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    public function isFunctionSupported(FunctionReflection $functionReflection): bool
+    {
+        return $functionReflection->getName() === 'shortcode_atts';
+    }
+
+    public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): Type
+    {
+        $argsCount = count($functionCall->args);
+        if ($argsCount === 0) {
```
szepeviktor, line 27:
```suggestion
        if ($functionCall->args === []) {
```

Is it okay for you?

monojp: Of course it is, you're the boss :) I'll update it later.

This is taken from another extension here. I just saw that the example in the phpstan docs do it without the assignment now, but you're way is fine too.

szepeviktor:
Thank you!
🚢 The release is on its way...

PR #43, opened 2021-05-01
Fix WP_Nav_Menu_Args

IanDelMar:
Adds the `container_aria_label` argument which was added in WP 5.5.0.


szepeviktor:
@IanDelMar Thank you!

PR #44, opened 2021-06-19
Correct the default return type of `get_posts()`

johnbillion:
When `get_posts()` is called and the `fields` argument is not included, the return type is `array<int,WP_Post>`.


szepeviktor:
> Passing anything else will return all fields (default) – an array of post objects.

https://developer.wordpress.org/reference/classes/wp_query/#return-fields-parameter

szepeviktor:
Thank you.

PR #46, opened 2021-06-19
Correct the return type of `get_taxonomies()` when `$output` is not a string

johnbillion:
The return type of `get_taxonomies()` is `array<int,string>` when the `$output` parameter is exactly `names` and `array<int,WP_Taxonomy>` in all other cases.

Ref: https://github.com/WordPress/wordpress-develop/blob/afee26086fea307c2a5c31c0e2018cb6acd598f7/src/wp-includes/taxonomy.php#L215

### DIFF
--- a/src/GetTaxonomiesDynamicFunctionReturnTypeExtension.php
+++ b/src/GetTaxonomiesDynamicFunctionReturnTypeExtension.php
@@ -11,7 +11,6 @@ namespace PHPStan\WordPress;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
-use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Type\Type;
 use PHPStan\Type\ArrayType;
 use PHPStan\Type\IntegerType;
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you again.

`if (! $argumentType instanceof ConstantStringType)`

Did you notice that static analysis can only process static values?
What should we do with a variable with indeterminable value as `$output`?

src/GetTaxonomiesDynamicFunctionReturnTypeExtension.php
```diff
@@ -33,27 +31,23 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
         $argumentType = $scope->getType($functionCall->args[1]->value);
 
         // When called with a non-string $output, return default return type
         if (! $argumentType instanceof ConstantStringType) {
-            return ParametersAcceptorSelector::selectFromArgs(
-                $scope,
-                $functionCall->args,
-                $functionReflection->getVariants()
-            )->getReturnType();
+            return new ArrayType(new StringType(), new ObjectType('WP_Taxonomy'));
```
szepeviktor, line 41: @johnbillion My problem is the following. When `get_taxonomies()` is given a variable PHPStan should let the return type unchanged.
What do you think?

szepeviktor (changes requested):

src/GetTaxonomiesDynamicFunctionReturnTypeExtension.php
szepeviktor: `@return array<string, string|\WP_Taxonomy>`

szepeviktor: Or is it `@return array<string|\WP_Taxonomy>`?

johnbillion: As fas as I can tell, `get_taxonomies()` returns `array<string,\WP_Taxonomy>` for any value of `$output` except for `'names'` when it returns `array<string,string>`. So it's always possible to determine the exact return type.

szepeviktor: > So it's always possible to determine the exact return type.

When static analysis cannot figure out the parameter type - as it is a variable, not a constant - we are not able to predict the return type. Do you agree?

PR #48, opened 2021-06-19
Add a return type extension for `WP_Query->query()` and `WP_Query->get_posts()`

johnbillion:
Return type extension for `WP_Query->query()` and `WP_Query->get_posts()`.

As `get_posts()` ultimately calls `WP_Query->query()` and `WP_Query->get_posts()`, they all share the same return type pattern. Not sure if you want to combine them somehow.

### DIFF
--- a/src/WPQueryDynamicMethodReturnTypeExtension.php
+++ b/src/WPQueryDynamicMethodReturnTypeExtension.php
@@ -72,10 +72,9 @@ class WPQueryDynamicMethodReturnTypeExtension implements \PHPStan\Type\DynamicMe
         }
 
         switch ($fields) {
+            case 'id=>parent':
             case 'ids':
                 return new ArrayType(new IntegerType(), new IntegerType());
-            case 'id=>parent':
-                return new ArrayType(new IntegerType(), new ObjectType('stdClass'));
             case 'all':
             default:
                 return new ArrayType(new IntegerType(), new ObjectType('WP_Post'));
--- a/src/WPQueryDynamicMethodReturnTypeExtension.php
+++ b/src/WPQueryDynamicMethodReturnTypeExtension.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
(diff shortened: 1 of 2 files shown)


src/WPQueryDynamicMethodReturnTypeExtension.php
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @see https://developer.wordpress.org/reference/classes/wp_query/#return-fields-parameter
+     */
+    public function getTypeFromMethodCall(MethodReflection $methodReflection, MethodCall $methodCall, Scope $scope): Type
+    {
+        // Called without arguments
+        if (count($methodCall->args) === 0) {
+            return new ArrayType(new IntegerType(), new ObjectType('WP_Post'));
+        }
```
szepeviktor, line 45: https://developer.wordpress.org/reference/classes/wp_query/get_posts/ has no parameters.
The `query` method has one.
What do you think?

szepeviktor (changes requested):

src/WPQueryDynamicMethodReturnTypeExtension.php
johnbillion: `$query = new WP_Query( $args ); $query->get_posts()` will cause the database query to run twice

swissspidy: Ah, right, meant another method then. What I wanted to point out is that args can be passed to the constructor.

johnbillion: Ah I see, yes. I need to find some time to come back to this PR.

szepeviktor: @johnbillion How about finishing this PR?

szepeviktor:
> closed this by deleting the head repository

@johnbillion What about this PR?

PR #49, opened 2021-06-19
Correct the return type for `get_posts()` with `id=>parent`

johnbillion:
Using `id=>parent` returns `array<int,int>`, not `array<int,stdClass>`.

Refs:

* https://github.com/johnbillion/wordpress-develop/blob/b6046636f015f6e1fac6b565f883fa9679b90a72/src/wp-includes/class-wp-query.php#L655-L656
* https://github.com/johnbillion/wordpress-develop/blob/b6046636f015f6e1fac6b565f883fa9679b90a72/src/wp-includes/class-wp-query.php#L2965-L2982


johnbillion:
I merged master in and this is passing now.

PR #51, opened 2021-07-14
Fix shortcode_atts extension

monojp:
Types were only inferred from the default attributes. This adds the string shortcode attributes as union values.

E.g.
```php
$atts = shortcode_atts(['foo' => null, 'bar' => 17], ['foo' => 'baz']);
```

resulted in `array('foo' => null, 'bar' => int)` previously while it now correctly results in `array('foo' => string|null, 'bar' => string|int)`

I was not a 100% sure about the implementation but apparently Ondrej approves ;) https://github.com/phpstan/phpstan/discussions/5303


szepeviktor:
This your own child!
https://github.com/szepeviktor/phpstan-wordpress/pull/36

szepeviktor:
You can upgrade :) https://github.com/szepeviktor/phpstan-wordpress/releases/tag/v0.7.7

monojp:
> This your own child!
Haha, I'm not sure if this is good or bad xD

thx!

PR #53, opened 2021-07-14
Fix phpstan problems

### LINKED ISSUE #45: Fix CS and static analysis
If you have a bit of time, please fix the tests.
See CI output for details.

Thank you!

monojp:
Tested manually via Composer VCS require

Fixes #45

### DIFF
--- a/src/WpThemeMagicPropertiesClassReflectionExtension.php
+++ b/src/WpThemeMagicPropertiesClassReflectionExtension.php
@@ -6,7 +6,7 @@
 
 declare(strict_types=1);
 
-namespace PHPStan\WordPress;
+namespace SzepeViktor\PHPStan\WordPress;
 
 use PHPStan\Reflection\ClassReflection;
 use PHPStan\Reflection\PropertyReflection;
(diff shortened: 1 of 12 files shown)


src/WpThemeMagicPropertiesClassReflectionExtension.php
```diff
@@ -31,6 +31,6 @@ public function hasProperty(ClassReflection $classReflection, string $propertyNa
     // phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter
     public function getProperty(ClassReflection $classReflection, string $propertyName): PropertyReflection
     {
-        return new DummyPropertyReflection();
+        return new DummyPropertyReflection(); // @phpstan-ignore-line ignore that this is not covered by BC
```
szepeviktor, line 34:
```suggestion
        // @phpstan-ignore-next-line This is not covered by BC.
        return new DummyPropertyReflection();
```

szepeviktor: What is the real solution instead of `DummyPropertyReflection`?

monojp: I don't know unfortunately, my knowledge of phpstan internals/extensions is unfortunately very limited too. Maybe it can be adapted in a follow-up

I force-pushed and adapted the ignore

szepeviktor:
This is a gift to me, Martin.
Thank you!

PR #56, opened 2021-10-04
Add a dynamic return type extension for `get_object_taxonomies()`

johnbillion:
`get_object_taxonomies()` function returns:

* `array<string,WP_Taxonomy>` when `$output` is `'objects'`
* `array<int,string>` otherwise

### DIFF
--- a/src/GetObjectTaxonomiesDynamicFunctionReturnTypeExtension.php
+++ b/src/GetObjectTaxonomiesDynamicFunctionReturnTypeExtension.php
@@ -1,0 +1,54 @@
+<?php
+
+/**
+ * Set return type of get_object_taxonomies().
+ */
+
+declare(strict_types=1);
+
+namespace SzepeViktor\PHPStan\WordPress;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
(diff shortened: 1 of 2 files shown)


src/GetObjectTaxonomiesDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,54 @@
...
+class GetObjectTaxonomiesDynamicFunctionReturnTypeExtension implements \PHPStan\Type\DynamicFunctionReturnTypeExtension
+{
+    public function isFunctionSupported(FunctionReflection $functionReflection): bool
+    {
+        return in_array($functionReflection->getName(), ['get_object_taxonomies'], true);
+    }
+
+    /**
+     * @see https://developer.wordpress.org/reference/functions/get_object_taxonomies/
+     */
```
szepeviktor, line 30:
```suggestion
    /**
     * @see https://developer.wordpress.org/reference/functions/get_object_taxonomies/
     *
     * phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter
     */
```

szepeviktor:
Thank you John!

PR #57, opened 2021-10-04
Add a return type extension for `mysql2date()`

johnbillion:
`mysql2date()` function returns:

* `false|int` when `$format` is `'G'` or `'U'`
* `false|string` otherwise


szepeviktor:
Thank you.
This seems to be perfect.

PR #66, opened 2021-11-01
Switch to PHPStan v1

johnbillion:
I've fixed the level to 8 because max uses the new level 9 which results in some errors with `mixed` which can be addressed at a later date.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
         "nikic/php-parser": "< 4.13.0",
         "php-parallel-lint/php-parallel-lint": "^1.1",
-        "phpstan/phpstan-strict-rules": "^1",
+        "phpstan/phpstan-strict-rules": "^1.0",
         "phpunit/phpunit": "^7 || ^9",
         "szepeviktor/phpcs-psr-12-neutron-hybrid-ruleset": "^0.6"
     },
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "require": {
         "php": "^7.1 || ^8.0",
         "php-stubs/wordpress-stubs": "^4.7 || ^5.0",
-        "phpstan/phpstan": "^0.12.26",
+        "phpstan/phpstan": "^1",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -13,15 +13,15 @@
     "require": {
         "php": "^7.1 || ^8.0",
         "php-stubs/wordpress-stubs": "^4.7 || ^5.0",
-        "phpstan/phpstan": "^0.12.26",
+        "phpstan/phpstan": "^1",
```
szepeviktor, line 16:
```suggestion
        "phpstan/phpstan": "^1.0",
```

composer.json
```diff
@@ -13,15 +13,15 @@
...
         "symfony/polyfill-php73": "^1.12.0"
     },
     "require-dev": {
         "composer/composer": "^1.10.23",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
         "nikic/php-parser": "< 4.13.0",
         "php-parallel-lint/php-parallel-lint": "^1.1",
-        "phpstan/phpstan-strict-rules": "^0.12",
+        "phpstan/phpstan-strict-rules": "^1",
```
szepeviktor, line 24:
```suggestion
        "phpstan/phpstan-strict-rules": "^1.0",
```

szepeviktor:
Should I release 1.0?

herndlm:
I can also test on a few internal wp plugins from us in ~30m

herndlm:
So I guess the things stated on https://github.com/nunomaduro/larastan/issues/964 are fine here? I can take another look at that as well just to be sure

szepeviktor:
Thank you. This extension is almost 100% `dynamicFunctionReturnTypeExtension`.

szepeviktor:
@johnbillion Thank you.
I wait for your PR-s before releasing v.1.0.

johnbillion:
@szepeviktor I think you can go ahead and release v1.0 👍  . I won't have time to work on those two PRs for a while.

herndlm:
Great, thx! As you expected everything is looking good. Tagging 1.0 makes sense too IMO. Looking forward to it :)

PR #67, opened 2021-11-02
Remove empty line before file header in example file

IanDelMar:
To be consistent with the other example files.


szepeviktor:
Thank you Ian!

PR #69, opened 2021-11-03
`apply_filters()` return type

johnbillion:
The return type of `apply_filters()` is indeterminate (therefore `mixed`). The return type is often wider than the type of the `$value` parameter being passed in, so it's not a generic function.

What we could do is introduce a dynamic function return type extension which reads the types from the docblock (if present) and treat those as the possible return types. This then places the burden on the developer to ensure the docblock is correct and subsequent code handles all possible return types.

Strictly, the only correct return type for `apply_filters()` is `mixed` because any filter can return a value of any type. In reality most code that uses `apply_filters()` assumes that it returns a value of the expected type. I need to give some more thought into whether treating the types in the docblock as absolute is more valuable than sticking with `mixed`.

### DIFF
--- a/src/ApplyFiltersDynamicFunctionReturnTypeExtension.php
+++ b/src/ApplyFiltersDynamicFunctionReturnTypeExtension.php
@@ -1,0 +1,73 @@
+<?php
+
+/**
+ * Set return type of apply_filters().
+ */
+
+declare(strict_types=1);
+
+namespace SzepeViktor\PHPStan\WordPress;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\PhpDoc\PhpDocStringResolver;
+use PHPStan\PhpDocParser\Lexer\Lexer;
+use PHPStan\PhpDocParser\Parser\ConstExprParser;
+use PHPStan\PhpDocParser\Parser\PhpDocParser;
+use PHPStan\PhpDocParser\Parser\TypeParser;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\Type;
+use PHPStan\Type\MixedType;
+
+class ApplyFiltersDynamicFunctionReturnTypeExtension implements \PHPStan\Type\DynamicFunctionReturnTypeExtension
+{
+    public function isFunctionSupported(FunctionReflection $functionReflection): bool
+    {
+        return in_array(
+            $functionReflection->getName(),
+            [
+                'apply_filters',
+            ],
+            true
+        );
+    }
+
+    // phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter
+    public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): Type
+    {
+        $default = new MixedType();
+
+        /** @var \PhpParser\Node\Expr\Assign|null */
+        $parent = $functionCall->getAttribute( 'parent' );
+
+        if (null === $parent) {
+            return $default;
+        }
+
+        // Fetch the docblock from the parent.
+        $comment = $parent->getDocComment();
+
+        if (null === $comment) {
+            return $default;
+        }
+
+        // Fetch the docblock contents and resolve it to a PhpDocNode.
+        $code = $comment->getText();
+        $docResolver = new PhpDocStringResolver( new Lexer(), new PhpDocParser( new TypeParser(), new ConstExprParser() ) );
+        $doc = $docResolver->resolve($code);
+
+        // Fetch the `@param` values from the docblock.
+        $params = $doc->getParamTagValues();
+
+        if (! $params) {
+            return $default;
+        }
+
+        // Fetch the `@param` types as a TypeNode.
+        $type = $params[0]->type;
+
+        // @TODO Fetch and return the Type that corresponds to the TypeNode.
+
+        return $default;
+    }
+}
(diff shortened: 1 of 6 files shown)


tests/data/apply_filters.php
```diff
@@ -0,0 +1,40 @@
...
+
+$value = apply_filters('filter','Hello, World');
+assertType('mixed', $value);
+
+/**
+ * Single type.
+ *
+ * @param string $foo Hello, World.
+ */
+$value = apply_filters('filter',$foo);
```
szepeviktor, line 19: The thing is value is usually a constant: a number or boolean. How to merry `@param` and a constant?

johnbillion: Good point, I added tests for this.

szepeviktor:
It is getting more exciting.

johnbillion:
Yeah it seems to be working well. I'm going to get more tests in for less common cases.

tests/data/apply_filters.php
```diff
@@ -0,0 +1,179 @@
...
+);
+assertType('string', $value);
+
+/**
+ * Single constant type.
+ *
+ * @param string $foo Hello, World.
+ */
+$value = apply_filters('filter','I am a string');
+assertType('string', $value);
```
szepeviktor, line 61: @johnbillion I think PHP language is not prepared for this: typehinting a string :)

Could your code handle this? `ConstantStringType` in PHPStan.

johnbillion: The return type isn't necessarily the same constant string though, otherwise `apply_filters()` would be a bit useless. Is that what you mean?

szepeviktor: > The return type

AFAIK you cannot set the return type of literal/constant values with `@param`. So we have to get the type from the constant thus the second parameter.

johnbillion: The fact that the parameter is constant doesn't matter. After passing through `apply_filters()` its type is set to whatever the `@param` tag says it is.

```php
add_filter( 'foo', fn() => 456 );

/**
 * @param int $number
 */
$number = apply_filters( 'foo', 123 );
```

The type of `$number` here is `int` not `123`.

szepeviktor: I think php/PHPStan/any reflection package won't connect `@param int $number` to `123`.
What do you think?

johnbillion: That's correct. That's why I've added the dynamic return type extension for `apply_filters()` which reads and "trusts" this WordPress-specific docblock placement.

I would like to write a custom rule at some point which validates the docblock by checking the types for each parameter match what's written in each `@param` tag, but that can be done separately from this dynamic return type extension.

szepeviktor: I see! `$params[0]->type`

johnbillion:
Okay we're getting closer now. The two remaining test failures are actual failures AFAICT so I will look at those later.

src/ApplyFiltersDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,106 @@
...
+            $scope->getFile(),
+            $scope->isInClass() ? $scope->getClassReflection()->getName() : null,
+            $scope->isInTrait() ? $scope->getTraitReflection()->getName() : null,
+            $scope->getFunctionName(),
+            $code
+        );
+
+        $nameScope = $resolvedPhpDoc->getNullableNameScope();
+
+        if (null === $nameScope) {
```
szepeviktor, line 99:
```suggestion
        if ($nameScope === null) {
```
Please tell Yoda to hunt another project!

src/ApplyFiltersDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        // Fetch the docblock contents and resolve it to a PhpDocNode.
+        $code = $comment->getText();
+        $doc = $this->phpDocStringResolver->resolve($code);
+
+        // Fetch the `@param` values from the docblock.
+        $params = $doc->getParamTagValues();
+
+        if (! $params) {
```
szepeviktor:
```suggestion
        if ($params === []) {
```
... I think ...

johnbillion: Already fixed :-)

src/ApplyFiltersDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,110 @@
...
+        $default = new MixedType();
+        $comment = self::getNullableNodeComment($functionCall);
+
+        if ($comment === null) {
+            return $default;
+        }
+
+        // Fetch the docblock contents and resolve it to a PhpDocNode.
+        $code = $comment->getText();
+        $doc = $this->phpDocStringResolver->resolve($code);
```
szepeviktor, line 61:
```suggestion
        /** @phpstan-ignore-next-line */
        $doc = $this->phpDocStringResolver->resolve($code);
```

> Calling PHPStan\PhpDoc\PhpDocStringResolver::resolve() is not covered by backward compatibility promise.

johnbillion: Yeah I need to look into whether there's an API method for this. I don't want to ignore it because it might break.

szepeviktor: PHPDoc parsing seem to be me as internal stuff...

szepeviktor:
Thank you!!

herndlm:
👍
love the commit messages btw. xD 
@szepeviktor you could squash merge next time to not have the intermediate commits on master? Just a suggestion, keep doing whatever you prefer 👍

PR #74, opened 2021-11-11
Test *ERROR*

szepeviktor:
@johnbillion

### DIFF
--- a/tests/data/_get_list_table.php
+++ b/tests/data/_get_list_table.php
@@ -12,3 +12,18 @@ assertType('Unknown_Table|false', _get_list_table('Unknown_Table'));
 
 // Unknown class name
 assertType('\WP_List_Table|false', _get_list_table(_GET['foo']));
+
+/**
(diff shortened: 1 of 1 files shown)


johnbillion:
Playground which works as expected: https://phpstan.org/r/d7b65195-af14-40cf-a216-9b70bdda21c0

szepeviktor:
@ondrejmirtes Could you help us why it fails in our simple PHPUnit test environment but it works in the playground?

ondrejmirtes:
You can't write test code like that. Functions need to be loaded. What I do most often is to write a class and test code inside that (because classes can be autoloaded automatically). If I need to use functions, I need to write test code in anonymous functions, like in this file: https://github.com/phpstan/phpstan-src/blob/master/tests/PHPStan/Analyser/data/generics.php

File in this format can be safely `require`'d.

szepeviktor:
Thank you for solving this.

PR #77, opened 2021-11-15
Docs for the `apply_filters()` handling

johnbillion:
It's not easy to explain this. Hopefully this is good enough.


szepeviktor:
Thank you.

PR #78, opened 2021-11-17
Hook docs rule

### LINKED ISSUE #65: Extend PHPStan scanning to cover `@param` tags of `apply_filters()` and `do_action()`
No idea if this is possible, I'm going to look into it and maybe ask on the PHPStan discussion forums.

When PHPStan encounters a call to `apply_filters()` or `do_action()` it should validate the types of the parameters passed to it according to the `@param` tags in the preceding docblock.

Example:

```php
/**
 * Filters the date formatted based on the locale.
 *
 * @param string       $date      Formatted date string.
 * @param string       $format    Format to display the date.
 * @param int          $timestamp Unix timestamp.
 * @param DateTimeZone $timezone  Timezone.
 */
$date = apply_filters( 'wp_date', $date, $format, $timestamp, $timezone );
```

Here, PHPStan should validate that `$date`, `$format`, `$timestamp`, and `$timezone` are of the correct types and trigger an error if not. This helps to ensure that the documentation for filters and actions is accurate.

It might be possible to achieve this by treating calls to `apply_filters()` and `do_action()` in the same manner as a closure. ~~As far as PHPStan is concerned, this code produces the same result:~~

```php
/**
 * Filters the date formatted based on the locale.
 *
 * @param string       $date      Formatted date string.
 * @param string       $format    Format to display the date.
 * @param int          $timestamp Unix timestamp.
 * @param DateTimeZone $timezone  Timezone.
 */
$date = function( string $date, string $format, int $timestamp, DateTimeZone $timezone ) : string {
    return $date;
}
```

johnbillion:
Implements validation of the `@param` tags in the optional docblock that precedes a call to `apply_filters()` or `do_action()`.

Fixes #65 .


szepeviktor:
Could you add a literal string test?

`apply_filters( 'filter', 'value1', $two );`

johnbillion:
There's lots left to do :-)

src/HookDocsRule.php
```diff
@@ -0,0 +1,141 @@
...
+        if ($numberOfParams !== $numberOfParamTags) {
+            $message = sprintf(
+                'Expected %1$d @param tags, found %2$d',
+                $numberOfParams,
+                $numberOfParamTags
+            );
+
+            if ($numberOfParams - 1 === $numberOfParamTags) {
+                foreach ($node->args as $param) {
+                    if ($param->value->name === 'this') {
```
szepeviktor, line 91: Could you add an incorrect test case for this to make sure this branch runs?

src/HookDocsRule.php
```diff
@@ -0,0 +1,141 @@
...
+                        // PHPStan does not detect param tags named `$this`, it skips the tag.
+                        // We can indirectly detect this by checking the actual parameter name,
+                        // and if one of them is `$this` assume that's the problem.
+                        $message = '@param tag must not be named $this. Choose a descriptive alias.';
```
szepeviktor, line 95:
```suggestion
                        $message = '@param tag must not be named $this. Choose a descriptive alias. For example $instance.';
```
How about this?

szepeviktor:
> Function is longer than 40 lines

This is highly critical. @johnbillion Do you have an idea what part/s could be extracted to a separate method?

johnbillion:
Not yet, but once I've got everything functionally covered I'll revisit it.

johnbillion:
@szepeviktor I think this is ready now, but there's a conflict between two of the PHPCS rulesets you're using.

If I don't import the `HookDocBlock` class then `complains. If I do import it then `SlevomatCodingStandard.Namespaces.UseFromSameNamespace.UseFromSameNamespace`.

I'll leave it up to you to decide what to do with that.

johnbillion:
Also I find many of the sniffs in the Slevomat coding standards that aren't in the "Functional" category to be unnecessarily prescriptive and quite the hindrance to being productive. In particular all the rules around fully qualified references which seem to go against the benefits of using namespaced code.

I'm all for coding standards, as you know, but when they introduce such a heavy burden it actually forms a roadblock to contributions. I would urge you to reconsider whether the full set of "Formatting" standards from that library are really necessary.

szepeviktor:
> In particular all the rules around fully qualified references

I follow a very simple rule in https://github.com/szepeviktor/phpcs-psr-12-neutron-hybrid-ruleset

1. Import all classes
2. Except for exceptions and PHPDoc blocks

What would you suggest instead?

PR #75, opened 2021-11-11
Test tests

szepeviktor:
Latest results: https://app.travis-ci.com/github/szepeviktor/phpstan-wordpress/jobs/547715235

PR #79, opened 2021-11-17
Configure CS for tests

herndlm:
Related to or part of #75. To make things easier I'm planning to create one PR per tool.

To allow simpler automagic fixing, I converted the inline phpcs arguments into a phpcs.xml.dist config file and added a `test:cs:fix` script. Most of the simple changes here were done with that. I also ignored some rules for the tests completely as they do not make much sense there IMO or are not workig as expected (e.g. `NeutronStandard.Functions.TypeHint` complains about a missing return type hint if `iterable` is used).

Let me know what you think

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,0 +1,28 @@
+<?xml version="1.0"?>
+<ruleset>
+	<file>bootstrap.php</file>
+	<file>src/</file>
+	<file>tests/</file>
+
+	<rule ref="PSR12NeutronRuleset">
+		<exclude name="Generic.Files.LineLength"/>
+		<exclude name="PEAR.Commenting.ClassComment"/>
+		<exclude name="PEAR.Commenting.FileComment"/>
+	</rule>
+
+	<rule ref="NeutronStandard.Functions.TypeHint">
+		<exclude-pattern>*/tests/*</exclude-pattern>
+	</rule>
+
+	<rule ref="NeutronStandard.Globals.DisallowGlobalFunctions">
+		<exclude-pattern>*/tests/*</exclude-pattern>
+	</rule>
+
+	<rule ref="WordPress.PHP.DiscouragedPHPFunctions">
+		<exclude-pattern>*/tests/*</exclude-pattern>
+	</rule>
+
+	<rule ref="WordPress.Security.NonceVerification">
+		<exclude-pattern>*/tests/*</exclude-pattern>
+	</rule>
+</ruleset>
--- a/tests/data/apply_filters.php
+++ b/tests/data/apply_filters.php
@@ -2,23 +2,25 @@
 
 declare(strict_types=1);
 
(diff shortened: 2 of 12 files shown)


herndlm:
I might have to change some things again when introducing phpstan. It might make also sense to wait for #78

phpcs.xml.dist
```diff
@@ -0,0 +1,28 @@
...
+	</rule>
+
+	<rule ref="WordPress.PHP.DiscouragedPHPFunctions">
+		<exclude-pattern>*/tests/*</exclude-pattern>
+	</rule>
+
+	<rule ref="WordPress.Security.NonceVerification">
+		<exclude-pattern>*/tests/*</exclude-pattern>
+	</rule>
+</ruleset>
```
szepeviktor, line 28: Oh noooooooooooooo :) You've killed my one-liner with a whole file.

tests/data/mysql2date.php
```diff
@@ -19,5 +19,5 @@
 assertType('int|string|false', mysql2date(get_option('date_format'), $time));
 
 // Unsupported types
-assertType('int|string|false', mysql2date(new \stdClass, $time));
+assertType('int|string|false', mysql2date(new \stdClass(), $time));
```
szepeviktor, line 22: Could you import this one?

szepeviktor (changes requested):

tests/data/apply_filters.php
```diff
@@ -105,102 +107,63 @@
...
  *
- * @param int[]        $max_image_size {
- *     An array of width and height values.
- *
- *     @type int $0 The maximum width in pixels.
- *     @type int $1 The maximum height in pixels.
+ * @param array<int> $max_image_size {
+ * An array of width and height values.
+ * @type int $0 The maximum width in pixels.
+ * @type int $1 The maximum height in pixels.
```
johnbillion, line 145: This shouldn't change because this test is to explicitly check this format of docblock. I think it makes more sense to ignore this file and its included files because it's likely to include code that intentionally does not adhere to the coding standards.

szepeviktor: > this format

Meaning WP core's own invention.

phpcs.xml.dist
herndlm: is that fine? xD

szepeviktor: Oh Yes!

tests/data/apply_filters.php
johnbillion: And the `array{ ... }` syntax used by PHPStan and Psalm is somebody else's invention, it's not part of FIG or phpDoc.

johnbillion: Regardless, it's irrelevant. This extension is for supporting syntax used in the WordPress ecosystem so its tests should reflect that.

herndlm: sure, makes sense to me. I missed that one after the auto-fix. adapted

szepeviktor:
Thank you for your work, Martin!

PR #80, opened 2021-11-17
Fix unit tests

herndlm:
Sorry, I broke them in #79 by renaming this file (phpunit did not execute those tests any more). This PR renames it again and removes the custom suffix config for phpunit. Unit tests should be running again now :)


szepeviktor:
No problemo!

PR #81, opened 2021-11-17
Configure phpstan and syntax check for tests

herndlm:
I had to configure the extensions itself so that phpstan knows about WP functions. So it kind of validates itself now? 🤯

This PR is related to #75 again.

@szepeviktor @johnbillion it looks like `apply_filters` is supporting more than 2 args, right? I ignored the errors that phpstan reported for that, maybe this should be fixed in a dedicated PR though?
I did not look yet, guess this is coming from the stubs?

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,4 @@
 includes:
-    - extension.neon
     - vendor/phpstan/phpstan/conf/bleedingEdge.neon
     - vendor/phpstan/phpstan-strict-rules/rules.neon
 parameters:
@@ -11,3 +10,5 @@ parameters:
         - bootstrap.php
         - src/
         - tests/
+    excludePaths:
+        - tests/data/*
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,4 +1,5 @@
 includes:
+    - extension.neon
     - vendor/phpstan/phpstan/conf/bleedingEdge.neon
     - vendor/phpstan/phpstan-strict-rules/rules.neon
 parameters:
@@ -9,3 +10,4 @@ parameters:
     paths:
         - bootstrap.php
         - src/
+        - tests/
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -11,4 +11,4 @@ parameters:
         - src/
         - tests/
     excludePaths:
-        - tests/data/*
+        - tests/data/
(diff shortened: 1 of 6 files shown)


szepeviktor:
>  `apply_filters` is supporting more than 2 args, right?

This is last popular error in WordPress stubs.
https://github.com/szepeviktor/phpstan-wordpress/blob/b76aae10d3a323cc9fbbb00c778289bf932f4e0c/example/phpstan.neon.dist#L11-L12

johnbillion:
The files in `tests/data` shouldn't be tested, they are files that are statically analyzed during the tests. They may also contain intentional errors as part of the testing.

tests/data/ApplyFiltersTestClass.php
```diff
@@ -38,7 +38,7 @@ public function myMethodWithParams(int $foo)
      *
      * @param int $foo Hello, World.
      */
-    public function anotherMethodWithParams(int $foo) // phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter
+    public function anotherMethodWithParams(int $foo): void // phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter
     {
```
szepeviktor, line 42:
```suggestion
    public function anotherMethodWithParams(int $foo): void
    {
        $useFoo = $foo;
```

herndlm:
Ok, I'd adapt it then to ignore `tests/data` and remove the "self-include" of the extension again

tests/data/apply_filters.php
```diff
@@ -135,6 +138,10 @@
 $value = apply_filters('filter', $foo);
 assertType("'aaa'|'bbb'", $value);
 
+$maxWidth = 17;
+$maxHeight = 19;
```
szepeviktor, line 142: @herndlm Why do we need to set these? They get assigned below.

herndlm:
Ok by excluding data again no errors show up right now ;)

phpstan.neon
```diff
@@ -9,3 +9,6 @@ parameters:
     paths:
         - bootstrap.php
         - src/
+        - tests/
+    excludePaths:
+        - tests/data/*
```
szepeviktor, line 14:
```suggestion
        - tests/data/
```

szepeviktor:
Could you add `tests/` to syntax check here?
https://github.com/szepeviktor/phpstan-wordpress/blob/b76aae10d3a323cc9fbbb00c778289bf932f4e0c/composer.json#L51

szepeviktor:
Thank you!
Great to have you on board.

PR #82, opened 2021-11-17
Add shortcode_atts tests

herndlm:
This thing is complex and debattable if it is correct enough, but I wanted to add some tests for the current state at least


szepeviktor:
Thanks!

PR #83, opened 2021-11-18
Comment out ignore rule for `load_plugin_textdomain()`

IanDelMar:
The `@param` type for the `$deprecated` and `$plugin_rel_path` arguments of `load_plugin_textdomain()` have been corrected in WP 5.2.

See https://github.com/WordPress/wordpress-develop/blob/5.2/src/wp-includes/l10n.php#L760-L777 vs. https://github.com/WordPress/wordpress-develop/blob/5.1/src/wp-includes/l10n.php#L760-L776.


szepeviktor:
Thank you, Ian.

PR #86, opened 2021-11-19
Introduce a dynamic return type extension for functions that include a `$wp_error` parameter

johnbillion:
Several functions in WordPress include a `$wp_error` parameter which causes the return type to vary depending on its boolean value. This PR introduces a dynamic return type extension which narrows the return types to either include or exclude `WP_Error` based on the value of this parameter.

Additionally, the default return types (when the value of `$wp_error` is unknown) are specified in order to narrow them and improve them. The docs in core aren't completely accurate.

I would like some more eyes on the return types to ensure that I've gotten them completely correct. Thanks!

---

**Functions covered:**

* [x] `wp_insert_link()`
* [x] `wp_insert_category()`
* [x] `wp_set_comment_status()`
* [x] `wp_update_comment()`
* [x] `wp_schedule_single_event()`
* [x] `wp_schedule_event()`
* [x] `wp_reschedule_event()`
* [x] `wp_unschedule_event()`
* [x] `wp_clear_scheduled_hook()`
* [x] `wp_unschedule_hook()`
* [x] `_set_cron_array()`
* [x] `wp_insert_post()`
* [x] `wp_update_post()`
* [x] `wp_insert_attachment()`

**Functions that will _not_ be covered in this PR:**

These functions are a mess and include conditional `never` so I'm not including them. We can tackle them at a later date.

* [x] `wp_allow_comment()`
* [x] `wp_new_comment()`


szepeviktor:
> Whether to return a WP_Error object on failure.

phpcs.xml.dist
```diff
@@ -4,6 +4,8 @@
 	<file>src/</file>
 	<file>tests/</file>
 
+	<exclude-pattern>*/tests/data/*</exclude-pattern>
```
szepeviktor, line 7: @johnbillion What went wrong with test data?

szepeviktor: I see!
Please insert a file comment.

```
/**
 * Test data for WPErrorParameterDynamicFunctionReturnTypeExtension.
 */
```

herndlm: We might be able to get rid of some of the tests exclusions for rules further down because of this in case you want to check
Otherwise I can and open a PR :)

szepeviktor: I'd like to keep all source code lines in good shape.

szepeviktor: I'm very sorry but this package is not a WordPress plugin. It is very dear to me.

szepeviktor: ...and yes `PSR12NeutronRuleset` is more than enough, it puts extra burden on everyone who follows it.

szepeviktor: You guys just contribute. I'll fix CS and extra shiny things in another PR.

herndlm: I don't want to derail the discussion here, but I think this would only be for the test data which has to cover some strange edge cases that can often not stick to specific coding standards. In general I'm also for strict standards

szepeviktor: You know... excluding a tiny part of the project from a specific check feels like _no blood will flow there_

johnbillion: These files are test data, they get statically analysed during the unit tests in order to verify the behaviour of the custom rules and return type extensions. They don't get executed.

PHPStan doesn't scan its own test data: https://github.com/phpstan/phpstan-src/blob/c1579f15c61db47950854c7bd17abfe177bca1d1/build/phpstan.neon#L24 . Some of its test data files contain intentional parse errors that are analysed during its tests to ensure it's handled correctly: https://github.com/phpstan/phpstan-src/blob/cb0ba9506c38d252b4b6b3bfe510e6f64bf04ba4/tests/PHPStan/Analyser/data/parse-error.php

szepeviktor: Thank you John. You cannot cool me down.
Please finish this PR and I'll spend time getting all file under the protective umbrella.

szepeviktor: I'm not able to maintain and publish this package without test.

johnbillion: I don't think I'm being very clear, sorry. It took me a while to understand what gets executed by PHP and what only gets statically read by PHPStan.

The files in `tests/data` are not _meant_ to be correct because they never get executed. They are a static representation of the potentially broken PHP code that people write, so that the tests in `tests/HookDocsRuleTest.php` and `tests/DynamicReturnTypeExtensionTest.php` can statically read them and assert that the code is handled correctly by this extension. The fact these files have a `.php` file extension is inconsequential. You can rename them `.txt` and everything works fine. Or you could rewrite all the tests to use strings of PHP for the test data instead of reading the PHP files.

Even the files that contain calls to `assertType(...)` don't get executed. They're statically read by PHPStan during the PHPUnit test run, which detects calls to `assertType(...)` and converts them into an runtime assertion (it's a bit weird to be honest).

* https://github.com/taptima/phpstan-custom/blob/master/phpstan.neon#L13
* https://github.com/phpstan/phpstan-doctrine/blob/master/phpstan.neon#L12
* https://github.com/mglaman/phpstan-drupal/blob/main/phpstan.neon#L15
* https://github.com/phpstan/phpstan-webmozart-assert/blob/master/phpstan.neon#L10
* etc

szepeviktor: Tests are mainly for me when I open these files and see thing I never want to see.

szepeviktor: Please wait for the next PR, I'll show you what I'm thinking of.

tests/data/wp_error_parameter.php
```diff
@@ -0,0 +1,308 @@
+<?php
```
szepeviktor, line 1:
```suggestion
<?php

/**
 * Test data for WPErrorParameterDynamicFunctionReturnTypeExtension.
 */
```

szepeviktor (approved):

szepeviktor:
I'm happy that `getArgs()` usage has been updated.
Thank you.

herndlm:
Viktor, you're so fast, I couldn't even take a second look xD

szepeviktor:
> I couldn't even take a second look xD

This is an open project! You can alter things after the merge.

PR #87, opened 2021-11-19
Use the same version of PHP Parser that PHPStan uses

johnbillion:
PHPStan >= 1.0 uses PHP Parser 4.13 which changed the handling of `Expr::$args`. This property should no longer be accessed directly; the `CallLike::getArgs()` method should be used instead. Release notes: https://github.com/nikic/PHP-Parser/releases/tag/v4.13.0

Here's where PHPStan made the change: https://github.com/phpstan/phpstan-src/commit/1c6ce5dfe4d16ad25ddef96a88541f7e44a3204f

The difference in behaviour is subtle and I only discovered it while working on #78. The `args` property is of type `array<Arg|VariadicPlaceholder>` but the return type of `CallLike::getArgs()` is `array<Arg>`. This means you can safely call methods on `Arg` that don't exist on `VariadicPlaceholder`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,6 @@
     "require-dev": {
         "composer/composer": "^1.10.23",
         "dealerdirect/phpcodesniffer-composer-installer": "^0.7",
-        "nikic/php-parser": "< 4.13.0",
         "php-parallel-lint/php-parallel-lint": "^1.1",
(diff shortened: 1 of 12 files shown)


szepeviktor:
Thanks.
You are right.

PR #88, opened 2021-11-20
PHPStan level 9

herndlm:
Not entirely sure if and where `FixWpStubs` is used though. Is it still needed?

### DIFF
--- a/src/Composer/FixWpStubs.php
+++ b/src/Composer/FixWpStubs.php
@@ -25,7 +25,7 @@ class FixWpStubs
         $io->write('Removing duplicate is_countable() ...');
 
         $vendorDir = $event->getComposer()->getConfig()->get('vendor-dir');
-        if (!is_string($vendorDir) || strlen($vendorDir) <= 0) {
+        if (!is_string($vendorDir) || $vendorDir === '') {
             $io->writeError('Could not determine Composer vendor dir.');
             return 12;
         }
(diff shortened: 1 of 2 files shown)


src/Composer/FixWpStubs.php
```diff
@@ -25,6 +25,11 @@ public static function php73Polyfill(Event $event): int
         $io->write('Removing duplicate is_countable() ...');
 
         $vendorDir = $event->getComposer()->getConfig()->get('vendor-dir');
+        if (!is_string($vendorDir) || strlen($vendorDir) <= 0) {
```
szepeviktor, line 28: Oh! 😱 negative-length strings :)

```suggestion
        if (!is_string($vendorDir) || strlen($vendorDir) !== '') {
```

herndlm: Those are the best strings. If you haven't worked with them, you haven't worked with strings!

szepeviktor: I have a special relations with strings.
https://gist.github.com/szepeviktor/6003fb6e2a35c0e5f9eb29c96623e4b7

szepeviktor:
Thank you for your work.

szepeviktor:
> Not entirely sure if and where `FixWpStubs` is used though. Is it still needed?

No, it is not needed.
This fix was for the pre-`php-stubs` era.

src/Composer/FixWpStubs.php
herndlm: I'm a bit scared now.

PR #91, opened 2022-02-14
Add a dynamic function return type extension for `term_exists()`

johnbillion:
The return type of `term_exists()` varies depending on its parameters. This adds a return type extension that covers this.


szepeviktor:
Thank you John.
Let's see CI and merge it.

PR #93, opened 2022-04-02
Introduce a dynamic return type extension for `has_filter()` and `has_action()`

johnbillion:
These functions return a boolean when `$callback` is false, otherwise an integer or false.


szepeviktor:
It seems simple.
Thank you!

PR #94, opened 2022-04-02
Introduce a dynamic return type extension for functions with an `$echo` or `$display` parameter

johnbillion:
These functions are [nearly] the worst. I've seen their non-existent return values used several times.

This covers all procedural functions with an `$echo` or `$display` parameter (all of which have a default value of `true`). There are some others which support an `$echo` argument inside an `$args` array, this doesn't cover those.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -60,6 +60,10 @@ services:
         tags:
             - phpstan.broker.dynamicFunctionReturnTypeExtension
     -
+        class: SzepeViktor\PHPStan\WordPress\EchoParameterDynamicFunctionReturnTypeExtension
+        tags:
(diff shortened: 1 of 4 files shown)


szepeviktor:
CI says green and I've taken a look at it.

szepeviktor:
@johnbillion Have you seen that recently more and more people use this package?

johnbillion:
No I haven't. Where are you looking at the stats?

szepeviktor:
The 3rd badge has a link to Packagist's stats: https://packagist.org/packages/szepeviktor/phpstan-wordpress/stats

Issue #30, opened 2020-12-11
Treat assertNotWPError() and assertWPError() as type-narrowing functions


johnbillion:
In the WordPress core test suite the following assertion methods are available:

* `$this->assertNotWPError()`
* `$this->assertWPError()`

It would be great if this extension treated these methods as type-narrowing functions, so the following code would not throw an error in PHPStan:

```php
$terms = wp_get_post_terms( $post_id, $taxonomy );
$this->assertNotWPError( $terms );
$this->assertCount( 0, $terms );
```

Currently this triggers the following error because `$terms` is assumed to be `array|WP_Error` on the last line.

```
[phpstan] Parameter #2 $haystack of method PHPUnit\Framework\Assert::assertCount() expects Countable|iterable, array|WP_Error given.
```

szepeviktor:
All right.
So you ask for `WP_UnitTestCase_Base::assertNotWPError()` never return `WP_Error` and `assertWPError()` should do the opposite?

szepeviktor:
Could you try installing `phpstan/phpstan-phpunit`?
https://github.com/phpstan/phpstan-phpunit#readme

johnbillion:
Yes the idea is that any parameter passed to `assertNotWPError()` cannot be a `WP_Error` in subsequent lines.

https://phpstan.org/writing-php-code/narrowing-types

I tried phpstan-phpunit but it had no effect, but I didn't play with it much.

szepeviktor:
I hope we can tell PHPStan that `assertNotWPError` is really `$this->assertNotInstanceOf( 'WP_Error', $actual, $message );` https://github.com/phpstan/phpstan-phpunit/blob/master/src/Type/PHPUnit/Assert/AssertTypeSpecifyingExtensionHelper.php

@ondrejmirtes Could you give us directions?

ondrejmirtes:
Yes, you need to write and register a similar extension that phpstan-phpunit has: https://github.com/phpstan/phpstan-phpunit/blob/master/src/Type/PHPUnit/Assert/AssertMethodTypeSpecifyingExtension.php

ondrejmirtes:
Some info about them also here: https://phpstan.org/developing-extensions/type-specifying-extensions

szepeviktor:
Thank you. Now we need those hands writing code.

szepeviktor:
@monojp Please advise.
What would you do in this case?

monojp:
[...]

PR #95, opened 2022-04-04
`is_wp_error()` rule

johnbillion:
This introduces a rule that triggers an error when the type of the value passed to `is_wp_error()` can never be a `WP_Error` (or can only be a `WP_Error`). It prevents code from incorrectly checking something that can never be an error (eg. checking the return value of `wp_insert_post()` when the `$wp_error` parameter is false).

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -77,6 +77,7 @@ services:
             - phpstan.broker.dynamicFunctionReturnTypeExtension
 rules:
     - SzepeViktor\PHPStan\WordPress\HookDocsRule
+    - SzepeViktor\PHPStan\WordPress\IsWpErrorRule
 parameters:
     bootstrapFiles:
         - %rootDir%/../../php-stubs/wordpress-stubs/wordpress-stubs.php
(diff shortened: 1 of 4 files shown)


szepeviktor:
Thank you.
Does it have to do something with  #30?

herndlm:
Hmm, shouldn't a normal return type extension be enough to let phpstans impossibletypechecker, or what it's called, handle this? 🤔

swissspidy:
Do you mean [`FunctionTypeSpecifyingExtension`](https://phpstan.org/developing-extensions/type-specifying-extensions)?

johnbillion:
There's already [a type specifying extension for `is_wp_error()`](https://github.com/szepeviktor/phpstan-wordpress/blob/master/src/IsWpErrorFunctionTypeSpecifyingExtension.php) but I didn't realise that should detect the impossible type situation. Wonder why it doesn't.

herndlm:
It might start to work if either `phpstan/phpstan-strict-rules` is used or `checkAlwaysTrueCheckTypeFunctionCall` https://phpstan.org/config-reference#checkalwaystruechecktypefunctioncall is set. But I'm not a 100% sure and activating those makes all checks realllllly strict. So maybe a dedicated rule is a good pragmatical approach after all :)

PR #96, opened 2022-04-22
Allow array like types in StringOrArrayDynamicFunctionReturnTypeExtension

Hug0-Drelon:
## Issue
Let's see a concrete example. In some cases, when calling `esc_sql($foo)`, PHPStan can interpret `$foo` as an `IntersectionType`. 
In such case, the returned value of `StringOrArrayDynamicFunctionReturnTypeExtension::getTypeFromFunctionCall()` is a `StringType` even if the provided `IntersectionType` is a mix of `ArrayType` and `NonEmptyArrayType`.
IMHO we should use `SomeType::isArray()` to check the provided argument to `esc_sql()` is indeed an array.

## Changes
- Use `isArray()` to check the argument type.
- Depending on the key type, return an associative array or an indexed array.
- Add unit tests for the `esc_sql()` function.

## To go further
Should I add some tests for `wp_slash` and `wp_unslash` too ?

### DIFF
--- a/src/StringOrArrayDynamicFunctionReturnTypeExtension.php
+++ b/src/StringOrArrayDynamicFunctionReturnTypeExtension.php
@@ -14,6 +14,7 @@ use PHPStan\Reflection\FunctionReflection;
 use PHPStan\Reflection\ParametersAcceptorSelector;
 use PHPStan\Type\StringType;
 use PHPStan\Type\ArrayType;
+use PHPStan\Type\IntegerType;
 use PHPStan\Type\Type;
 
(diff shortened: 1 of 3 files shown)


szepeviktor:
Thank you @Hug0-Drelon 

@johnbillion Could it be that there were no tests for `esc_sql`?

johnbillion:
Yeah `esc_sql()` return type extension was added well before there were any tests

szepeviktor:
> Should I add some tests for `wp_slash` and `wp_unslash` too ?

Thank you for your work.
You may send as many PR-s as you like.

PR #98, opened 2022-04-26
Add PHPStan bleeding edge compatibility

johnbillion:
It looks like the hook docblock detection broke in PHPStan bleeding edge in 1.6 due to removal of the `parent` attribute: https://phpstan.org/blog/preprocessing-ast-for-custom-rules .

The `getNullableNodeComment()` method needs to be updated so it doesn't use the `parent` node attribute.


szepeviktor:
Sound like _greek_ to me.

johnbillion:
Yeah the AST processing code is wild

johnbillion:
I think I know why the parsing on PHP < 7.4 isn't working, I'll take a look tomorrow.

szepeviktor:
> I think I know why the parsing on PHP < 7.4 isn't working

> phpstan/phpstan                                 1.4.10

No 1.6 :(

https://github.com/phpstan/phpstan/blob/1.6.1/composer.json#L6

szepeviktor:
For the PHP error in PHPCS on PHP 8.1: https://github.com/WordPress/WordPress-Coding-Standards/issues/2035#issuecomment-1077737225 👀

szepeviktor:
Still fails on PHP 7.2
```diff
--- Expected
+++ Actual
@@ @@
-'string'
+'PHPUnit\Framework\MockObject\Stub\ReturnStub'
```
https://github.com/szepeviktor/phpstan-wordpress/blob/master/tests/data/apply_filters.php#L84-L90

herndlm:
I started debugging this, not sure yet why it happens but the `returnValue` helper functions from here seems to be clashing with https://github.com/sebastianbergmann/phpunit/blob/main/src/Framework/Assert/Functions.php#L2676 which is also global. 2 possible fixes that I can think of
- add the same namespace to `test/functions.php` and drop the global use from the test files
- rename `returnValue`

I can open a PR against your branch John, if you like. or anyone of you feel free and go ahead and just adapt it, if that's faster :)

herndlm:
sorry for the PR noise, @johnbillion checkout https://github.com/johnbillion/phpstan-wordpress/pull/1 if you have time. that should fix it

szepeviktor:
@johnbillion Is it my turn now?

johnbillion:
Sure 🍸

Issue #99, opened 2022-04-28
Performance issues


herndlm:
I noticed that this extension increases both CPU and memory resources quite a bit. Sorry, I did not have time to profile it, but the "problems" seem to be there since 1.0.0 at least already.

I have quite a small plugin that uses the latest phpstan, phpstan-webmozart-assert extension and phpstan-wordpress extension.

Test case running `vendor/bin/phpstan --debug --verbose analyse`
With this extension: ~17s using ~310M memory
Without this extension:~4s using ~110M memory

Could be something really simple that we need to improve here, who knows. I can profile it soon hopefully. I see slowness in one specific file, so I'll need to inspect that one in particular.

herndlm:
I thought about that one as well, but assumed it was added very recently (after 1.0.0). I noticed that all versions since 1.0.0 are more or less affected. Previous ones are just a tiny bit faster. Must be something older I assume

herndlm:
OK, I tool a beer home with me from the office fridge I'd like to drink. I'll promise I'll only do that if I either fix this or at least know what _exactly_ is causing it. Wish me luck :)

szepeviktor:
Try commenting out this line see seconds and megabytes!
https://github.com/szepeviktor/phpstan-wordpress/blob/00fb9ff3b8199408ead0624be7f74f50e05d935a/extension.neon#L83
(there will be 1 zillion errors)

herndlm:
yep, you're absolute right. this "stub file" which is used as bootstrap file is a bit over 4M big and the reason for that apparently. It looks also like it got worse with the latest phpstan versions, e.g. if I go back I can see that almost every minor increased resources a bit :/
1.0.0: 10s/225M, 1.6.0: 14s/305M
1.0.0 (without stubs): 2s/70M,1.6.0 (without stubs): 3s/85M

herndlm:
fyi @ondrejmirtes, summary: apparently huge stub/bootstrap files make analysis slow and hog memory. splitting them into smaller files seems to be fine or better at least. does it make sense to look into phpstan itself improving handling of such big (~ 4M) stub/bootstrap files?

herndlm:
and another thing that seems to be improving the situation extremely is instead of specifying the huge stubs file as bootstrap file, directly specifying a wordpress-checkout via `scanDirectories`. That was really simple and gave me the best results so far. but how can this be done user friendly for people using this extension 🤔

ondrejmirtes:
[...]

PR #100, opened 2022-04-28
Switch from boostrap stubs to scanned wordpress dir

herndlm:
Closes https://github.com/szepeviktor/phpstan-wordpress/issues/99
Just playing around with an improvement idea.

had to temporarily limit the phpstan upper version for testing because currently 1.6.0 seems to be not fully compatible yet (HookDocsRuleTest).

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Please see [WooCommerce Stubs](https://github.com/php-stubs/woocommerce-stubs)
 ### What this extension does
 
 - Makes it possible to run PHPStan on WordPress plugins and themes
-- Loads [`php-stubs/wordpress-stubs`](https://github.com/php-stubs/wordpress-stubs) package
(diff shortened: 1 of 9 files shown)


herndlm:
I think this would work and improves performance quite a bit. The only downsides I see so far are
- Viktors stub file is not used anymore :/
- the WordPress checkout is bigger than the stub file (~60M or so)

herndlm:
fyi CI was green here, but as Ondrej mentioned this would need bleedingEdge and I think it is preferred to fix stub/bootstrap parsing performance, see issue for details

herndlm:
good idea, I'll look into the core problem in phpstan itself first and think about this again then

szepeviktor:
I think 101% of code developed for WordPress is written in PHP4-style, so they could use the speedup of `scanDirectories:`.

PR #101, opened 2022-05-02
Avoid namespace-clashing of test helpers

herndlm:
Should help move https://github.com/szepeviktor/phpstan-wordpress/pull/98 forward, see https://github.com/szepeviktor/phpstan-wordpress/pull/98#issuecomment-1115253979


herndlm:
This is expected to fail, it should be green with PHPStan 1.5.x or together with https://github.com/szepeviktor/phpstan-wordpress/pull/98

herndlm:
Ah this is messy, I'll just open the PR against John's branch..

PR #104, opened 2022-05-11
Dynamically include WordPress stubs

herndlm:
Since we have similar problems here as composer bins have, I came up with the same solution basically, e.g. see https://github.com/sebastianbergmann/phpunit/blob/9.5.20/phpunit#L66, but   I'm not a 100%  happy with it, maybe there's a better solution.

Refs https://github.com/szepeviktor/phpstan-wordpress/pull/103


phpcs.xml.dist
```diff
@@ -3,6 +3,7 @@
     <file>bootstrap.php</file>
     <file>src/</file>
     <file>tests/</file>
+    <file>wordpress-stubs.php</file>
```
szepeviktor, line 6: Please do not check the stubs. Stubs are having a hard life.

szepeviktor (changes requested):

phpcs.xml.dist
herndlm: it's now just a bootstrap file like bootstrap.php :)

szepeviktor: In other words: I'm not able to format stubs.

szepeviktor: My mistake.

wordpress-stubs.php
```diff
@@ -0,0 +1,12 @@
...
+declare(strict_types=1);
+
+foreach ([__DIR__ . '/../../php-stubs/wordpress-stubs/wordpress-stubs.php', __DIR__ . '/vendor/php-stubs/wordpress-stubs/wordpress-stubs.php'] as $file) {
+    if (!file_exists($file)) {
+        continue;
+    }
+
+    require_once $file;
+    break;
+}
```
szepeviktor, line 12:
```suggestion
// This file locates and loads WordPress core stubs.

foreach ([dirname(__DIR__, 2) . '/php-stubs/wordpress-stubs/wordpress-stubs.php', __DIR__ . '/vendor/php-stubs/wordpress-stubs/wordpress-stubs.php'] as $file) {
    if (file_exists($file)) {
        require_once $file;
        break;
    }
}
```

szepeviktor:
I'm in a state of _unusable_.
Please wait for my recovery!

herndlm:
> I'm in a state of _unusable_. Please wait for my recovery!

Victor you sometimes talk in such a cryptic way :D But sure, take however long you need!
I still hope that we come up with a better solution.

szepeviktor:
I'm not able to grasp this problem. (I had a nightmare in the morning)

szepeviktor:
Martin! Good news. Final solution is here.

Please only put unconditional loading of `__DIR__ . '/vendor/php-stubs/wordpress-stubs/wordpress-stubs.php'` in `wordpress-stubs.php`.
No other change is needed 🎉

PR #105, opened 2022-05-11
Fix tests by only touching tests

szepeviktor:
@herndlm How shocked are you??


herndlm:
Viktor the magician strikes again!

Very creative, I like the fact that there is no special handling outside of tests needed that can potentially fail

bart-jaskulski:
@szepeviktor Great solution and I think, the closest to mimic end user environment. Thank you!

szepeviktor:
Thank you.

Issue #106, opened 2022-05-18
Add type checking for hook callbacks in `add_action` and `add_filter`


johnbillion:
I'm planning on working on some rules that check the use of `add_action()` and `add_filter()` when the action or filter is from WordPress core. [The wp-hooks library](https://github.com/johnbillion/wp-hooks) facilitates this, and [the wp-hooks-generator library](https://github.com/johnbillion/wp-hooks-generator) facilitates this for hooks in third party plugins and themes (although I'm only concentrating on the WordPress core hooks for now).

I'll open a PR once I have something working.

---

When I call `add_action( 'foo', $callback )` or `add_filter( 'foo', $callback )`:

* The signature of the callback function should be checked against the parameters that get passed to the hook
* The number used in the `$accepted_args` parameter should match the number of parameters that the callback accepts
* A filter should not be used as an action, nor vice versa
* The callback for `add_filter` must return a value with a type which is compatible with the type passed to it
* The callback for `add_action` must return void

Notes:

* It's fine for the callback function to accept fewer parameters than are passed to the hook (but see point above about `$accepted_args`)
* There is no requirement to have a docblock on the callback because it's the job of PHPStan to figure out parameter types
* we could also look up the literal hook name, eg `plugin_action_links_{$plugin_file}` and see if we get a match for dynamic hooks

szepeviktor:
High expectations!
Thank you.

herndlm:
This was completed, wasn't it? 🎉

szepeviktor:
I do not follow the development :)

johnbillion:
This was only partially completed, I'll open a follow-up issue with the details.

menno-ll:
Also one extra thing, which is more an opinion i would like to start a little discussion about.

When creating a callback function for an action, with this new set of rules, a return type of `void` is forced for that function.
However, because of it beeing an action, nothing would happen if you would let the callback function return a value.
So why do I bring this up?
It's because in some occasions, you would like to use the callback function in another place in your project. And there, you actually want to use the return variable.
And this is now not possible.

An example below:

```php

add_action( 'save_post', 'set_custom_meta' );
function set_custom_meta( int $post_id ): bool {
    return (bool) update_post_meta( $post_id, 'some_meta', 'some_value' );
}

function some_function_somewhere_else_in_the_website(): void {
    $meta_updated = set_custom_meta( 1 );
    if ( $meta_updated ) {
        // Do something
    } else {
        // Do something
[...]

PR #107, opened 2022-06-01
Add type checking for hook callbacks in `add_action` and `add_filter`

johnbillion:
This implements part of #106.

I'm learning more about PHPStan internals as I work on this, so it's going to take some time 😄  . Once everything's working and covered by tests then I'll refactor it.

Applicable to all hooks:

* [X] The number used in the `$accepted_args` parameter should match the number of parameters that the callback accepts
* [x] The callback for `add_filter` must return a value
* [x] The callback for `add_action` must return void

~~Applicable when the hook signature is known (eg. a WordPress core hook):~~

* [ ] ~~The signature of the callback function should be checked against the parameters that get passed to the hook~~
* [ ] ~~A filter should not be used as an action, nor vice versa~~
* [ ] ~~The callback for `add_filter` must return a value with a type which is compatible with the type passed to it~~

I've decided to implement the WordPress core hook handling in a separate PR because it relies on [the wp-hooks API package that I'm writing](https://github.com/wp-hooks/api) and it's not ready yet.


src/HookCallbackRule.php
```diff
@@ -0,0 +1,129 @@
...
+        $hookNameValue = null;
+
+        if ($hookNameType instanceof ConstantStringType) {
+            $hookNameValue = $hookNameType->getValue();
+        }
+
+        $callbackType = $scope->getType($callbackArg->value);
+
+        // If the callback is not valid, bail out and let PHPStan handle the error:
+        if ($callbackType->isCallable()->no()) {
```
herndlm, line 88: For trinary logic like that most likely you want to check `!..->yes()` to also exclude maybe I guess?

szepeviktor:
This PR brings a fantastic feature to `phpstan-wordpress`.

johnbillion:
I'm stuck on `isCallable()->yes()` not returning the expected value in some cases. Opened https://github.com/phpstan/phpstan/discussions/7715 to ask for help.

szepeviktor:
I love when one find an upstream bug during working.

src/HookCallbackRule.php
```diff
@@ -0,0 +1,197 @@
...
+
+        // If we don't have enough arguments, let PHPStan handle the error:
+        if (count($args) < 2) {
+            return [];
+        }
+
+        $callbackType = $scope->getType($args[1]->value);
+
+        // If the callback is not valid, let PHPStan handle the error:
+        if (! $callbackType->isCallable()->yes()) {
```
szepeviktor, line 78: @johnbillion When it fails is it a "no" or a "maybe"?

johnbillion: Figured it out, the `isCallable()` check is working as expected now, I have some other return type handling left to do.

johnbillion:
We're all green ✅ but I want to add some more tests in.

szepeviktor:
kép

szepeviktor (approved):
I'm waiting for our local PHPStan specialist.
@herndlm

johnbillion:
@szepeviktor @herndlm Would be great to get this in

src/HookCallbackRule.php
```diff
@@ -0,0 +1,218 @@
...
+    private const SUPPORTED_FUNCTIONS = [
+        'add_filter',
+        'add_action',
+    ];
+
+    /** @var \PHPStan\Rules\RuleLevelHelper */
+    protected $ruleLevelHelper;
+
+    /** @var \PHPStan\Analyser\Scope */
+    protected $currentScope;
```
herndlm, line 37: really more like a personal opinion/style kind of thing I guess, but I would rather pass the Scope through the methods instead of assigning it for later internal usage. That also would make this rule immutable by removing the state I think. PHPStan itself is also doing it like that.

src/HookCallbackRule.php
```diff
@@ -0,0 +1,218 @@
...
+        return $acceptedArgs;
+    }
+
+    protected function validateParamCount(ParametersAcceptor $callbackAcceptor, ?Arg $acceptedArgsParam): void
+    {
+        $acceptedArgs = $this->getAcceptedArgs($acceptedArgsParam);
+
+        if ($acceptedArgs === null) {
+            return;
+        }
```
herndlm, line 121: just for my understanding - null means basically invalid type for acceptArgs and we don't check it, correct?

src/HookCallbackRule.php
```diff
@@ -0,0 +1,218 @@
...
+        );
+    }
+
+    protected function validateActionReturnType(ParametersAcceptor $callbackAcceptor): void
+    {
+        $acceptedType = $callbackAcceptor->getReturnType();
+        $accepted = $this->ruleLevelHelper->accepts(
+            new VoidType(),
+            $acceptedType,
+            true
```
herndlm, line 190: I really don't know much about the RuleLevelHelper yet, but I see in PHPStan internals that the `$strictTypes` arg  is normally set  to `Scope::isDeclareStrictTypes()`. Does it make sense to also not enforce strict types here, but set it to whatever the current file is using too?

src/HookCallbackRule.php
```diff
@@ -0,0 +1,218 @@
...
+            $expectedParametersMessage,
+            $acceptedArgs
+        );
+    }
+
+    protected function validateActionReturnType(ParametersAcceptor $callbackAcceptor): void
+    {
+        $acceptedType = $callbackAcceptor->getReturnType();
+        $accepted = $this->ruleLevelHelper->accepts(
+            new VoidType(),
```
herndlm, line 188: I never even used it myself yet, but does it make sense maybe to also add tests for  the `never` return type? Not sure what we'd expect, e.g. if PHPStan or this rule is going to complain that the callable never returns / always exits 🤔

herndlm (commented):
Sorry for taking so long to take a look at this, I was a bit frightened away in the beginning by the big diff, but this seems to be mostly coming from tests :)

I'm not a specialist for PHPStan Rules tbh, you might already know them better yourself. But I checked the rule and added a couple of comments, in general it looks well-made to me and very useful! I did not check the test thoroughly, you surely know the correct WP function behaviour there better than me :)

src/HookCallbackRule.php
johnbillion: Correct, if it's not a constant integer then it gets ignored. PHPStan will trigger an error. This test covers that: https://github.com/szepeviktor/phpstan-wordpress/pull/107/files#diff-56e889cfab3a17dd1d1fbc7b198fcf3130ca973b4a7fd3342c696d29df13a96dR130-R132

src/HookCallbackRule.php
johnbillion: Cheers. I'll leave it like this for now because we've already got other rules in the repo that set this as a property and I'd rather go for consistency. We can always refactor the rules later.

src/HookCallbackRule.php
johnbillion: To be honest I haven't figured this out myself, hence why I went with `true`. The behaviour of PHPStan itself differs depending on whether `declare(strict_types=1)` is present in the file being analysed, but I'm not sure how to vary the behaviour of an extension based on that. I might need to ask in the discussion forums.

src/HookCallbackRule.php
johnbillion: I'll take a look!

johnbillion: Ah I've already got tests for this! https://github.com/szepeviktor/phpstan-wordpress/pull/107/files#diff-56e889cfab3a17dd1d1fbc7b198fcf3130ca973b4a7fd3342c696d29df13a96dR199-R213.

Unfortunately it is acceptable for both filters and actions to exit because WP YOLO. I'm not sure how `VoidType` relates to `NeverType` but this is working as expected and the tests pass, so I think we're all good here.

herndlm: oh, ok. I missed those tests

szepeviktor:
Thank you! and Thank you!

szepeviktor:
@johnbillion @herndlm We have a bunch of problems.
https://app.travis-ci.com/github/szepeviktor/phpstan-wordpress/builds/257619024
Are they coming from this PR?

szepeviktor:
Or PHPStan, or WP stubs?

szepeviktor:
Last successful test back in September was also with WP 6.0.1

> Locking php-stubs/wordpress-stubs (v6.0.1)

herndlm:
there are still the \FQCN problems there was another issue about. I had an easy solution in mind, but I think you didn't like it because it involved a phpstan min version bump. but I'll check that one out quickly.

and the rest are theme related type changes? could be stubs?

szepeviktor:
> there are still the \FQCN problems

Let's close our eyes for this type of problems!

johnbillion:
Yeah might be related to #118 or #119 :(

szepeviktor:
All right guys!
Here comes a new release ...
🪨

herndlm:
oh, this was actually one thing, the FQCN change was causing the WP_Theme extension class to not do anything anymore. should be fixed with https://github.com/szepeviktor/phpstan-wordpress/pull/122

swissspidy:
@johnbillion How well is this supposed to work yet? I am getting lots of "Filter callback return statement is missing" false positives.

Example:

https://github.com/GoogleForCreators/web-stories-wp/blob/c3b5eec59b0bb79dc096260f5c9614cffc3b3778/includes/Admin/Admin.php#L75C45-L78

The top three lines are getting flagged because those methods don't have a return type (just in PHPDoc).
The last one has a PHP return type and is not flagged.

Looks like it doesn't take `@return` doc blocks into account at all.

johnbillion:
It's supposed to be perfect, beautiful, flawless, delightful, and kind, but it probably needs some tweaking. Can you open a new issue please?

szepeviktor:
FYI @swissspidy There is no `string|mixed` type as mixed include Everything :)

Please open that issue with the first problem 🚀

herndlm:
thx again for this, renovatebot is just updating this in our projects, CI still green so far 🤞 🎉

PR #108, opened 2022-06-01
Add support for WP 6.0 stubs.

ZebulanStanphill:
Since WordPress (and likewise `php-stubs/wordpress-stubs`) doesn't follow SemVer, I assume this one-liner is all that's necessary.


szepeviktor:
Thank you Zebulan!!

PR #110, opened 2022-06-21
Add `WpParseUrlFunctionDynamicReturnTypeExtension`

swissspidy:
I noticed that PHPStan has a custom extension to handle `parse_url()` return types, but no such extension exists for `wp_parse_url()`.

This PR straight up copies `ParseUrlFunctionDynamicReturnTypeExtension` from PHPStan so one gets more accurate types for `wp_parse_url()` than just "mixed"

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -76,6 +76,10 @@ services:
         tags:
             - phpstan.broker.dynamicFunctionReturnTypeExtension
     -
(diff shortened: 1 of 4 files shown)


szepeviktor:
Thank you Pascal!!

Could it be that that this code is meant to be run on PHP 8.1?

kép

PR #111, opened 2022-07-22
Add the_title() to the echo parameter dynamic return type extension

IanDelMar:
This PR adds the [the_title()](https://developer.wordpress.org/reference/functions/the_title/) as a supported function in the echo parameter dynamic return type extension.


szepeviktor:
All checks are green 🍏 so I should merge it :)

Thank you @IanDelMar!

PR #112, opened 2022-07-23
Add dynamic return type extension for WP_Theme::get()

IanDelMar:
This PR adds a dynamic return type extension for the get method of WP_Theme.

### DIFF
--- a/src/WpThemeGetDynamicMethodReturnTypeExtension.php
+++ b/src/WpThemeGetDynamicMethodReturnTypeExtension.php
@@ -1,0 +1,78 @@
+<?php
+
+// phpcs:disable SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter
+
+/**
+ * Set return type of WP_Theme::get().
+ */
+
+declare(strict_types=1);
+
+namespace SzepeViktor\PHPStan\WordPress;
+
+use PhpParser\Node\Expr\MethodCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\MethodReflection;
+use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+
+class WpThemeGetDynamicMethodReturnTypeExtension implements \PHPStan\Type\DynamicMethodReturnTypeExtension
+{
+
+    /**
+     * File headers.
+     *
+     * @var array<string,string>
+     */
+    private static $headers = [
+        'Name' => 'Theme Name',
+        'ThemeURI' => 'Theme URI',
+        'Description' => 'Description',
+        'Author' => 'Author',
+        'AuthorURI' => 'Author URI',
+        'Version' => 'Version',
+        'Template' => 'Template',
+        'Status' => 'Status',
+        'Tags' => 'Tags',
+        'TextDomain' => 'Text Domain',
+        'DomainPath' => 'Domain Path',
+        'RequiresWP' => 'Requires at least',
+        'RequiresPHP' => 'Requires PHP',
+    ];
+
+    public function getClass(): string
+    {
+        return '\WP_Theme';
+    }
+
+    public function isMethodSupported(MethodReflection $methodReflection): bool
+    {
+        return $methodReflection->getName() === 'get';
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


src/WpThemeGetDynamicMethodReturnTypeExtension.php
```diff
@@ -0,0 +1,78 @@
...
+use PHPStan\Type\Constant\ConstantBooleanType;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+
+class WpThemeGetDynamicMethodReturnTypeExtension implements \PHPStan\Type\DynamicMethodReturnTypeExtension
+{
+
```
szepeviktor, line 25:
```suggestion
{
```

szepeviktor:
Thank you.

src/WpThemeGetDynamicMethodReturnTypeExtension.php
IanDelMar: What is this about? The commit suggestions button is disabled and telling me that "This suggestion is invalid because no changes were made to the code."

szepeviktor: I don't know how to remove a line with suggestion tool ...

src/WpThemeGetDynamicMethodReturnTypeExtension.php
```diff
@@ -0,0 +1,78 @@
...
+    /**
```
szepeviktor, line 26:
```suggestion
class WpThemeGetDynamicMethodReturnTypeExtension implements \PHPStan\Type\DynamicMethodReturnTypeExtension
{
    /**
```

szepeviktor:
Could it be that the values in the `header` property is unused?
Is it possible to use a list of strings?

src/WpThemeGetDynamicMethodReturnTypeExtension.php
```diff
@@ -0,0 +1,77 @@
...
+
+        if (!$argumentType instanceof ConstantStringType) {
+            return new ConstantBooleanType(false);
+        }
+
+        if ($argumentType->getValue() === 'Tags') {
+            return new ArrayType(new IntegerType(), new StringType());
+        }
+
+        if (isset(self::$headers[$argumentType->getValue()])) {
```
szepeviktor, line 71: Something like
```suggestion
        if (in_array(self::$headers, $argumentType->getValue(), true)) {
```

IanDelMar: `in_array()` checks the values, but we need to check the keys.

szepeviktor: :) We don't use values right now.
So we need a simple list of strings. Could you do that?

src/WpThemeGetDynamicMethodReturnTypeExtension.php
```diff
@@ -0,0 +1,77 @@
...
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+
+class WpThemeGetDynamicMethodReturnTypeExtension implements \PHPStan\Type\DynamicMethodReturnTypeExtension
+{
+    /**
+     * File headers.
+     *
+     * @var array<string,string>
```
szepeviktor, line 29:
```suggestion
     * @var list<string>
```

szepeviktor:
It look much better now!

Thank you.

tests/data/wp_theme_get.php
```diff
@@ -0,0 +1,24 @@
...
+assertType('string', $theme->get('AuthorURI'));
+assertType('string', $theme->get('Version'));
+assertType('string', $theme->get('Template'));
+assertType('string', $theme->get('Status'));
+assertType('array<int, string>', $theme->get('Tags'));
+assertType('string', $theme->get('TextDomain'));
+assertType('string', $theme->get('DomainPath'));
+assertType('string', $theme->get('RequiresWP'));
+assertType('false', $theme->get('something'));
+assertType('false', $theme->get($_GET['something']));
```
szepeviktor, line 24: This needs to be changed too.

szepeviktor:
I'm so exxxcited!

IanDelMar:
Thank you for assisting!

herndlm:
Clean and simple, looking good 👍

szepeviktor:
OFF [Ich lerne Deutsch](https://simonorsi.hu/), auch am Samstag.

PR #113, opened 2022-07-26
Improve dynamic return type extension for get_post()

IanDelMar:
This PR improves the dynamic return type extension for `get_post()` and `get_page_by_path()` by
1. removing `null` if the first argument `$post` is of type `WP_Post`,
2. dropping to return the default type hints in favor of always returning `WP_Post|null` as long as the second argument `$output` is not `ARRAY_A` or `ARRAY_N`.

Reasoning:

1. If `$post` is an instance of `WP_Post` the variable `_post` is set to `$post` (see [here](https://github.com/WordPress/WordPress/blob/87464289c387757943d07e8474d8dab61ba7b6f2/wp-includes/post.php#L995-L997)) and will never fail the existence check (see [here](https://github.com/WordPress/WordPress/blob/87464289c387757943d07e8474d8dab61ba7b6f2/wp-includes/post.php#L1010-L1012)) and hence will never be of type `null`.
2. Both `get_post()` and `get_page_by_path()` only check `$output` against `ARRAY_A` and `ARRAY_N`. If `$output` is set to something different `get_post()` and `get_page_by_path()` will always return an instance of `WP_Post` or `null` irrespective of the type and the value of `$output` (see [here](https://github.com/WordPress/WordPress/blob/87464289c387757943d07e8474d8dab61ba7b6f2/wp-includes/post.php#L1016-L1022)). Hence there's also no need to type check `$output`.

### DIFF
--- a/src/GetPostDynamicFunctionReturnTypeExtension.php
+++ b/src/GetPostDynamicFunctionReturnTypeExtension.php
@@ -52,7 +52,9 @@ class GetPostDynamicFunctionReturnTypeExtension implements \PHPStan\Type\Dynamic
             // When called with an $output that isn't a constant string, return default return type
             if (! $scope->getType($args[1]->value) instanceof ConstantStringType) {
                 return TypeCombinator::union(
-                    $objectType, $associativeArrayType, $numericArrayType
+                    $objectType,
+                    $associativeArrayType,
+                    $numericArrayType
                 );
             }
             if ($args[1]->value instanceof ConstFetch) {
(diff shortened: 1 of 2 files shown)


szepeviktor:
Thank you @IanDelMar!

@johnbillion Could you review it please? You represent "WordPress" here :)

johnbillion:
Nice work. This all seems good except the following case:

``php
assertType('WP_Post|null', get_page_by_path('page/path', $_GET['foo']));
```

With an unknown value for `$output` it could have a value that matches `ARRAY_A` or `ARRAY_N`, therefore the type should remain as `array|\WP_Post|null`.

szepeviktor:
@IanDelMar Could you make that change?

IanDelMar:
@johnbillion You're absolutely right. Could you have a look at the changes I made to the PR? Thank you.
@szepeviktor Done.

johnbillion (approved):
😰 😄

PR #114, opened 2022-08-31
Switch to stubFiles

johnbillion:
This doesn't work, but I don't know why.

Ref: https://phpstan.org/user-guide/stub-files

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -91,8 +91,9 @@ rules:
     - SzepeViktor\PHPStan\WordPress\HookDocsRule
     - SzepeViktor\PHPStan\WordPress\IsWpErrorRule
 parameters:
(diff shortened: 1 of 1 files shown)


szepeviktor:
> I don't know why

Good for you!
This is a very dirty subject. Source of bad feelings.

You are better off closing this PR.

szepeviktor:
Actually my brain's self-defense kicked in and I forgot the technical details.

Good for me too! :)
I do not want to participate in a mission impossible.

herndlm:
Or could it be the discovering symbols related part from what John linked? Feels related to what Viktor mentioned with autoloading too.

Using a debugger or messing around with phpstan-src should tell what errors are actually behind the error type.

PR #117, opened 2022-09-22
Add `get_sites()` dynamic function return type extension

### LINKED ISSUE #116: Add dynamic return type extension for `get_sites()`
The `get_sites()` function in WordPress returns of `WP_Site[]`, `int[]` when 'fields' is set to 'ids', or `int` when 'count' is passed as a query var.

swissspidy:
Also added tests for `get_post()` while at it

Fixes #116


src/GetSitesDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,77 @@
...
+use PHPStan\Type\IntegerType;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\Constant\ConstantStringType;
+
+class GetSitesDynamicFunctionReturnTypeExtension implements \PHPStan\Type\DynamicFunctionReturnTypeExtension
+{
+    public function isFunctionSupported(FunctionReflection $functionReflection): bool
+    {
+        return in_array($functionReflection->getName(), ['get_sites'], true);
```
szepeviktor, line 25:
```suggestion
        return $functionReflection->getName() === 'get_sites';
```

src/GetSitesDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        switch ($fields) {
+            case 'count':
+                return new IntegerType();
+            case 'ids':
+                return new ArrayType(new IntegerType(), new IntegerType());
+            default:
+                return new ArrayType(new IntegerType(), new ObjectType('WP_Site'));
+        }
```
szepeviktor, line 75:
```suggestion
        switch ($fields ?? null) {
            case 'count':
                return new IntegerType();
            case 'ids':
                return new ArrayType(new IntegerType(), new IntegerType());
            default:
                return new ArrayType(new IntegerType(), new ObjectType('WP_Site'));
        }
```

Does it work?

szepeviktor: or put `$fields = '';` somewhere at the top ...

src/GetSitesDynamicFunctionReturnTypeExtension.php
swissspidy: Wanna change that in `GetPostsDynamicFunctionReturnTypeExtension` as well? That's where I copied this from

src/GetSitesDynamicFunctionReturnTypeExtension.php
szepeviktor: No `?? null` please.
Initialize `$fields` instead.

src/GetSitesDynamicFunctionReturnTypeExtension.php
szepeviktor: `in_array` is for possible other functions with the same type problem.
Do as you with please!

szepeviktor:
There are no more road blocks.

Thank you Pascal!

szepeviktor:
https://github.com/szepeviktor/phpstan-wordpress/releases/tag/v1.1.3

PR #118, opened 2022-10-06
Use `::class` to get the fully qualified name of the class

IanDelMar:
From PHPStan 1.8.6 `return '\WP_Theme';` no longer works. PHPStan skips the dynamic return type when analyzing an instance of `WP_Theme`.

### DIFF
--- a/src/WpThemeGetDynamicMethodReturnTypeExtension.php
+++ b/src/WpThemeGetDynamicMethodReturnTypeExtension.php
@@ -46,7 +46,7 @@ class WpThemeGetDynamicMethodReturnTypeExtension implements \PHPStan\Type\Dynami
 
     public function getClass(): string
     {
-        return '\WP_Theme';
+        return \WP_Theme::class;
     }
 
     public function isMethodSupported(MethodReflection $methodReflection): bool


IanDelMar:
The 3 failures from the checks are not related to this PR.

herndlm:
The change makes sense I guess, but I wonder why this behaviour changed in phpstan and if it is a BC break. You don't know what in phpstan broke it do you?

szepeviktor:
In string form the must not be leading backslash.

szepeviktor:
Sorry guys, this is too hard for me.
Please see https://github.com/phpstan/phpstan/issues/8118

szepeviktor:
This is how FQCN should be written https://github.com/szepeviktor/phpstan-wordpress/pull/119

FQCN only for docblocks and exception.

I'm hesitant with `extends` and `implements`. I do not see a reason except name collision.

szepeviktor:
You have to know that this repo was not meant to be for PHP code.
Only for stubs+some shell code.

szepeviktor:
https://github.com/phpstan/phpstan-src/pull/1758

How to go on?
PHPStan v1.8.7 outputs class names without a backslash.
Requiring this latest version is not the nicest thing in the world.

szepeviktor:
> the easiest thing

I let other people choose that.

IanDelMar:
@herndlm I don't know what causes the issue or whether one of the sources mentioned by @szepeviktor is the cause. 

@szepeviktor The changes proposed in https://github.com/szepeviktor/phpstan-wordpress/pull/119 do not resolve the issue addressed by this PR. Using https://github.com/szepeviktor/phpstan-wordpress/pull/119 makes PHPStan "ignore" the dynamic method return type in v1.8.5, v1.8.6 and v1.8.7.

IanDelMar:
Using 
```php
public function getClass(): string
{
        return \WP_Theme::class;
}
```
also matches the dynamic return type example  from the [docs](https://phpstan.org/developing-extensions/dynamic-return-type-extensions):
```php
public function getClass(): string
{
	return \Doctrine\ORM\EntityManager::class;
}

public function isMethodSupported(MethodReflection $methodReflection): bool
{
	return $methodReflection->getName() === 'merge';
}

public function getTypeFromMethodCall(
	MethodReflection $methodReflection,
	MethodCall $methodCall,
	Scope $scope
): Type
{ ... }
```

szepeviktor:
It is hard to be a repo owner!

szepeviktor:
@johnbillion Could you jump in a resolve the https://github.com/szepeviktor/phpstan-wordpress/pull/118#issuecomment-1269671946 wich a single click?

herndlm:
> How does the `WP_Theme` class exist during static analysis? Surely the change should be to just remove the leading slash from the string.

👍 + phpstan version bump, no? Otherwise people come with old phpstan versions and complain the slash is missing I guess. If I understood the problem..

szepeviktor:
> does it make sense to test with lowest and highest deps in CI

@herndlm I how a full-blow software project looks like :) This is definitely not like that.

szepeviktor:
Does anyone know how to get phpstan version in tests?

IanDelMar:
> Does anyone know how to get phpstan version in `DynamicMethodReturnTypeExtension`?

What about `\PHPStan\Internal\ComposerHelper::getPhpStanVersion()`?

szepeviktor:
@IanDelMar Could you please add `"phpstan/phpstan": "<1.8.6"` to your Composer configuration?
I need time to digest all these.

IanDelMar:
> @IanDelMar Could you please add `"phpstan/phpstan": "< 1.8.6"` to your Composer configuration? I need time to digest all these.

Sure. Take your time.

szepeviktor:
@IanDelMar Done in master :)

szepeviktor:
Thank you.

PR #119, opened 2022-10-06
Always import classes

szepeviktor:


szepeviktor:
@herndlm Does this PR do any harm?

herndlm:
sorry, only saw this now. this is only CS, right? no, I personally prefer to import everything, phpstan does too. you could enforce this via e.g. https://github.com/slevomat/coding-standard/blob/master/doc/namespaces.md#slevomatcodingstandardnamespacesreferenceusednamesonly- which we also use

szepeviktor:
Thank you for your response.

There is a simple rule: import everything + FQCN for docblocks and exception.

PR #122, opened 2022-11-09
Fix FQCN types

herndlm:
Fixes the FQCN types by requiring PHPStan 1.8.7 where the behaviour was fixed causing a BC break here.

Refs: https://github.com/phpstan/phpstan/issues/8118
> Likely an **expected** side effect of this change https://github.com/phpstan/phpstan-src/pull/1758

Also if you check the changed tests files - each of them has other assertions where the backslash was missing, so this was already inconsistent and this is the best way to "fix" it here and move forward IMO.

PR #123, opened 2022-11-09
Run tests on lowest deps also

szepeviktor:
@herndlm Here we go 🐎 

From https://github.com/szepeviktor/phpstan-wordpress/pull/122#issuecomment-1309363724

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ jobs:
       php: "7.4"
       install:
         - "composer update --no-interaction --prefer-lowest"
-        - "composer update --no-interaction --with-all-dependencies szepeviktor/phpcs-psr-12-neutron-hybrid-ruleset"
+        - "composer update --no-interaction --with-all-dependencies szepeviktor/phpcs-psr-12-neutron-hybrid-ruleset 'php-stubs/wordpress-stubs:^6.1'"
 
 cache:
   directories:
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ jobs:
       php: "7.4"
       install:
         - "composer update --no-interaction --prefer-lowest"
+        - "composer update --no-interaction --with-all-dependencies szepeviktor/phpcs-psr-12-neutron-hybrid-ruleset"
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


szepeviktor:
>   - Locking wp-coding-standards/wpcs (2.1.0)

herndlm:
yep, the phpcs errors can be fixed by upgrading the non direct dependency `wp-coding-standards/wpcs` from 2.1.0 to 2.3.0  but then some other CS errors show up, but it's a step forward

szepeviktor:
> upgrading the non direct dependency `wp-coding-standards/wpcs`

Done.

herndlm:
`composer require --dev szepeviktor/phpcs-psr-12-neutron-hybrid-ruleset:^0.6.1 automattic/phpcs-neutron-standard:^1.7` fixes all phpcs errors but one

szepeviktor:
Let's just ignore CS.
I'm not willing to fiddle with my CS for hours.

herndlm:
the dev deps in general are not important at all for the lowest deps IMO. annoying that they get downgraded

herndlm:
a good pragmatical solution might be to run the lowest tests with just e.g. `composer update --prefer-lowest --prefer-stable php-stubs/wordpress-stubs phpstan/phpstan`

herndlm:
looks like you're determined to fix it 🎉
for the stub-dependency I think I can build you some conditional tests. e.g. one mysql2date for WP 5 and one for WP 6 or so, or just lock them to WP 6

szepeviktor:
@herndlm Dev deps fixed!

How to handle inconsistent WP types?

szepeviktor:
AFAIK WP 6.0 introduced meaningful types. Up to 5.0 types were mainly ignored.

szepeviktor:
You can't cook a tasty soup of mud!!

herndlm:
hehe, I wanted to suggest to wrap some tests maybe in something like
```php
if (InstalledVersions::satisfies(new VersionParser(), 'php-stubs/wordpress-stubs', '^5.0')) {
            yield from $this->gatherAssertTypes(__DIR__ . '/data/wp_error_parameter.php');
        }
```

but locking the stubs should be fine, right. lower phpstan version is the most interesting thing most likely :)

herndlm:
I think this makes sense, thx Viktor

szepeviktor:
Today's German lesson had a new word: _Alptraum_ - that is something I do not want!

PR #124, opened 2022-11-10
Add `AssertWpErrorTypeSpecifyingExtension` / `AssertNotWpErrorTypeSpecifyingExtension`

swissspidy:
Might need a bit of help with the implementation & tests, but I think this should cover the basics.

Fixes #30

PR #127, opened 2022-11-10
Fix AssertWpErrorTypeSpecifying extensions

herndlm:
Viktor you were right, this was autoloader related. PHPStan needs to be able to cleanly autoload this class in order to fire the extensions. This is my first hacky approach to fix that autoloading with some kind of stub class copied into the tests folder.

The rest are just tiny fixes for the extensions themselves.

Refs: https://github.com/szepeviktor/phpstan-wordpress/pull/124 CC @swissspidy

### DIFF
--- a/src/AssertWpErrorTypeSpecifyingExtension.php
+++ b/src/AssertWpErrorTypeSpecifyingExtension.php
@@ -28,21 +28,17 @@ class AssertWpErrorTypeSpecifyingExtension implements \PHPStan\Type\MethodTypeSp
 
     public function isMethodSupported(MethodReflection $methodReflection, MethodCall $node, TypeSpecifierContext $context): bool
     {
-        return strtolower($methodReflection->getName()) === 'assertWPError'
+        return strtolower($methodReflection->getName()) === 'assertwperror'
             && isset($node->args[0])
-            && !$context->null();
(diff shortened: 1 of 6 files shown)


src/AssertWpErrorTypeSpecifyingExtension.php
```diff
@@ -28,21 +28,17 @@ public function getClass(): string
 
     public function isMethodSupported(MethodReflection $methodReflection, MethodCall $node, TypeSpecifierContext $context): bool
     {
-        return strtolower($methodReflection->getName()) === 'assertWPError'
+        return strtolower($methodReflection->getName()) === 'assertwperror'
```
szepeviktor, line 31: Is it lower or camelCase?

szepeviktor:
Only 1 thing (2 occurrences)
```diff
- $yes = $yes;
+ $yes = $_GET['yes'];
```
Would it be possible?

herndlm:
Should be possible. The type is specified via phpdoc and overwritten anyways. I'm just not in front of the computer anymore, so feel free to adapt if you want

herndlm:
Also the diff I produced on one test file is not needed and could be reverted. Sorry, I was in a hurry and missed that..

PR #129, opened 2022-11-11
Increase the memory limit during unit tests

johnbillion:
Running `composer test:phpunit` locally results in an out of memory error hidden behind the mysterious `255` error code from PHPUnit. This fixes that by bumping the limit to 1G.

Yes the underlying memory usage should be fixed, no I don't have time to look into it at the moment.

cc @swissspidy

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
         "test:cs": "phpcs",
         "test:cs:fix": "phpcbf",
         "test:phpstan": "phpstan analyze",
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you for this PR.

Developer environment setting could be in another project, it is normal.
I'd like to keep this project clean from environment stuff.

szepeviktor:
The default value in Debian for `memory_limit` on CLI is -1 (unlimited).

PR #130, opened 2022-11-11
Hook callback false positives for `mixed` return type

### LINKED ISSUE #125: False positives in `add_filter()` type checking
Re-sharing from https://github.com/szepeviktor/phpstan-wordpress/pull/107#issuecomment-1309987538

The new rules check that callbacks added for `add_filter` actually return a value.

In my initial testing it doesn't always work, the rule only seems to check the return type declaration, but is ignoring any `@return` docblocks.

Example:

https://github.com/GoogleForCreators/web-stories-wp/blob/c3b5eec59b0bb79dc096260f5c9614cffc3b3778/includes/Admin/Admin.php#L75-L78

The top three lines are getting flagged because those methods don't have a return type (just in PHPDoc).
The last one has a PHP return type and is not flagged.

johnbillion:
Fixes #125

Seems the problem is ultimately down to `mixed` return types.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
         "test:cs": "phpcs",
         "test:cs:fix": "phpcbf",
         "test:phpstan": "phpstan analyze",
-        "test:phpunit": "phpunit -d memory_limit=1G",
+        "test:phpunit": "phpunit",
         "test:syntax": "parallel-lint bootstrap.php src/ tests/"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
         "test:cs": "phpcs",
         "test:cs:fix": "phpcbf",
         "test:phpstan": "phpstan analyze",
-        "test:phpunit": "phpunit",
+        "test:phpunit": "phpunit -d memory_limit=1G",
         "test:syntax": "parallel-lint bootstrap.php src/ tests/"
     }
 }
--- a/tests/data/hook-callback.php
+++ b/tests/data/hook-callback.php
@@ -95,9 +95,6 @@ add_filter('filter', '__return_false', 10, 2);
 // Action callback returns false but should not return anything (with incorrect number of accepted args).
 add_action('action', '__return_false', 10, 2);
 
-// Filter callback return statement is missing.
-add_filter('filter', __NAMESPACE__ . '\\no_return_value_untyped');
-
 // Action callback returns mixed but should not return anything.
 add_action('action', __NAMESPACE__ . '\\return_value_mixed');
 
@@ -148,7 +145,7 @@ add_filter('filter', function(array $classes): array {
 
 // Callback function with no declared return type.
 add_action('action', __NAMESPACE__ . '\\return_value_untyped');
-add_action('action', __NAMESPACE__ . '\\return_value_implicit_mixed');
+add_filter('filter', __NAMESPACE__ . '\\no_return_value_untyped');
 
 /**
  * Correct usage:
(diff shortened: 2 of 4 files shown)


herndlm:
also looking at this right now :)

yeah, the problem is that a `mixed` return type currently cannot be distinguished from a missing one (which is also mixed). I think you could check mixed explicitness (the one should be explicit while the other is implicit). That means some `instanceof MixedType` checks which are not optimal, but I can't think of anything else right now ..

herndlm:
oh and btw another test case for implicit mixed is simply something like the following
```php
function return_value_mixed_implicit( $value ) {
    return $value;
}
```
if you put `return 123;` there it's most likely not mixed anymore

tests/data/hook-callback.php
```diff
@@ -240,6 +248,33 @@ function return_value_typed() : int {
...
+}
+
+function return_value_implicit_mixed( $value ) {
+    return $value;
+}
+
+/**
+ * Return type documented as a union that includes mixed.
+ *
+ * @return int|mixed
```
szepeviktor, line 268: I'm sorry to report there is no `*|mixed` type.
I see it all over the Internet. Please keep this project clean.

herndlm: it will be mixed, yeah. I guess this was meant as regression test for the reported bug :)

johnbillion: Yeah this is a useless union type. I assume that under the hood it's treated as `mixed` by PHPStan so it should be handled the same as `mixed`. This was indeed added as a regression test for the bug reported by Pascal.

tests/data/hook-callback.php
```diff
@@ -240,6 +251,33 @@ function return_value_typed() : int {
...
+function return_value_mixed() {
+    return 123;
+}
+
+function return_value_implicit_mixed( $value ) {
+    return $value;
+}
+
+/**
+ * Return type documented as a union that includes mixed.
```
szepeviktor, line 266:
```suggestion
 * Return type documented as a union that includes mixed.
 *
 * This is added as a regression test for a bug.
```

herndlm:
Fyi using the following patch seems to greatly improve it already for me. But we can still not differentiate between no return at all and mixed implicit return. Type-wise it makes sense, since PHPStan is not looking at the function body but only the definition and unfortunately I don't know if there's another way of getting the info if there's a `return` inside basically. Sticking with the types might be safer

```patch
     protected function validateFilterReturnType(ParametersAcceptor $callbackAcceptor): void
     {
         $returnType = $callbackAcceptor->getReturnType();
-        $isVoidSuperType = $returnType->isSuperTypeOf(new VoidType());

-        if (! $isVoidSuperType->yes()) {
+        if (
+            ($returnType instanceof MixedType && $returnType->isExplicitMixed())
+            || !$returnType->isSuperTypeOf(new VoidType())->yes()
+        ) {
             return;
         }
```

johnbillion:
I think we got a bit distracted by implicit mixed, which isn't the responsibility of this library to figure out. If your callback uses implicit mixed then PHPStan flags the callback as having a missing return type ([these tests](https://github.com/szepeviktor/phpstan-wordpress/blob/4800e7cce402a344c0e6bd412f31b1bc7ad433bd/tests/data/hook-callback.php#L101-L148) are in place to ensure this library doesn't trigger an error in that situation).

szepeviktor:
You paint CI feedback green, I'll merge it right away.

herndlm:
good point. In that case - does it make sense to also use RuleLevelHelper then, as you do with add_action? Or turn the supertype check around like `(new VoidType())->isSuperTypeOf($returnType)` to report when something is surely a void return?

johnbillion:
Okay I removed the implicit mixed handling (which is how it should be), added support for explicit mixed handling, and CI is all green.

composer.json
```diff
@@ -56,7 +56,7 @@
         "test:cs": "phpcs",
         "test:cs:fix": "phpcbf",
         "test:phpstan": "phpstan analyze",
-        "test:phpunit": "phpunit",
+        "test:phpunit": "phpunit -d memory_limit=1G",
```
szepeviktor, line 59:
```suggestion
        "test:phpunit": "phpunit",
```

szepeviktor:
Thank you!

I hope every user will be happy.

herndlm:
Nice, thank you!

PR #131, opened 2022-11-11
Fix mixed return type for filter callbacks

Hug0-Drelon:
## Issue:
I get the following error for a filter callback with a mixed return type:
```
Filter callback return statement is missing.
```

This occurs with a callback like this:
```php
/**
 * Does things.
 *
 * @param mixed  $param  Some value.
 * @return mixed
 */
function do_things( $param ) {
    // Do things
    return $param
}
```

IMHO this is due to this [line](https://github.com/szepeviktor/phpstan-wordpress/blob/e1503589f0a0ed4950c467558d99fbad108a21ca/src/HookCallbackRule.php#L208) where we get `MixedType` which seems to be a super type of `VoidType` (maybe because it include `NullType`.

## Proposal:
- Add a special case for `MixedType`.
- Add a test for a callback with mixed return type.

### DIFF
--- a/src/HookCallbackRule.php
+++ b/src/HookCallbackRule.php
@@ -19,6 +19,7 @@ use PHPStan\Rules\RuleLevelHelper;
 use PHPStan\Type\Constant\ConstantIntegerType;
 use PHPStan\Type\VerbosityLevel;
 use PHPStan\Type\VoidType;
+use PHPStan\Type\MixedType;
 
 /**
  * @implements \PHPStan\Rules\Rule<\PhpParser\Node\Expr\FuncCall>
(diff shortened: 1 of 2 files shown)


szepeviktor:
@herndlm @johnbillion Please help @Hug0-Drelon with his issue.

szepeviktor:
I think we are working on this in  #130

Hug0-Drelon:
Oops, I should have look into previous PRs... Sorry.
Also, I have a hard time to run PHPUnit, for unknown reason it doesn't work locally...

szepeviktor:
> I have a hard time to run PHPUnit

Set `memory_limit` in your php.ini

Hug0-Drelon:
Useless since https://github.com/szepeviktor/phpstan-wordpress/pull/130 is merged and fixed the issue.

PR #139, opened 2023-02-04
Add dynamic return type for `get_permalink()`/`get_the_permalink()`

IanDelMar:
Adds dynamic function return type extension for `get_permalink()` and `get_the_permalink()`.

### DIFF
--- a/src/GetPermalinkDynamicFunctionReturnTypeExtension.php
+++ b/src/GetPermalinkDynamicFunctionReturnTypeExtension.php
@@ -1,0 +1,57 @@
+<?php
+
+/**
+ * Set return type of get_permalink() and get_the_permalink().
+ */
+
+declare(strict_types=1);
+
+namespace SzepeViktor\PHPStan\WordPress;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Reflection\ParametersAcceptorSelector;
+use PHPStan\Type\ObjectType;
+use PHPStan\Type\StringType;
+use PHPStan\Type\Type;
+
+class GetPermalinkDynamicFunctionReturnTypeExtension implements \PHPStan\Type\DynamicFunctionReturnTypeExtension
+{
+    public function isFunctionSupported(FunctionReflection $functionReflection): bool
+    {
+        return in_array(
+            $functionReflection->getName(),
+            [
+                'get_permalink',
+                'get_the_permalink',
+            ],
+            true
+        );
+    }
+
+    /**
+     * @see https://developer.wordpress.org/reference/functions/get_permalink/
+     * @see https://developer.wordpress.org/reference/functions/get_the_permalink/
+     */
+    public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): Type
+    {
+        $args = $functionCall->getArgs();
+
+        if (count($args) > 0) {
+            $type = $scope->getType($args[0]->value);
+
+            // Called with a WP_Post instance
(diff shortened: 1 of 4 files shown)


src/GetPermalinkDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,57 @@
...
+            $type = $scope->getType($args[0]->value);
+
+            // Called with a WP_Post instance
+            if ($type instanceof ObjectType && $type->isInstanceOf('WP_Post')->yes()) {
+                return new StringType();
+            }
+        }
+
+        // When called without arguments or with a $type that isn't a WP_Post instance, return default return type
+        return ParametersAcceptorSelector::selectFromArgs(
```
herndlm, line 51: It should be also possible to just return `null` instead if you want to simplify this

herndlm (commented):
Also small idea for an improvement - what do you think of letting it return false in case the arg is not an object AND not an integer? That could help find some false usages of those functions :)

src/GetPermalinkDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * @see https://developer.wordpress.org/reference/functions/get_permalink/
+     * @see https://developer.wordpress.org/reference/functions/get_the_permalink/
+     */
+    public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): Type
+    {
+        $args = $functionCall->getArgs();
+
+        if (count($args) > 0) {
```
szepeviktor, line 41:
```suggestion
        if (count($args) !== 0) {
```

szepeviktor:
@IanDelMar Thank you for working on a Saturday! 📅

IanDelMar:
> Also small idea for an improvement - what do you think of letting it return false in case the arg is not an object AND not an integer? That could help find some false usages of those functions :)

PHPStan knows that the first argument has to be of type `int` or `WP_Post` because of the stubs file.

IanDelMar:
> Also small idea for an improvement - what do you think of letting it return false in case the arg is not an object AND not an integer? That could help find some false usages of those functions :)

PHPStan knows that the first argument must be of type `int` or `WP_Post` because of the stubs file. So there is an error if you use a non integer, non WP_Post argument. Using a string type argument gives `Parameter #1 $post of function get_permalink expects int|WP_Post, string given.`

Using the return value as an argument for a function that expects a string, e.g. `str_replace( 'a string', 'a new string', get_permalink( 'not expected string' ) )` gives `Parameter #3 $subject of function str_replace expects array|string, string|false given.`

But maybe it is helpful for some users to have `Parameter #3 $subject of function str_replace expects array|string, false given.` instead.

src/GetPermalinkDynamicFunctionReturnTypeExtension.php
szepeviktor: @IanDelMar Could you try using `null` here?

IanDelMar: Can you give an example when `get_permalink()` returns `null`?

IanDelMar: Ah, now I see. You mean returning `null` is equivalent to returning the default type(s)?

IanDelMar: > It should be also possible to just return `null` instead if you want to simplify this

True. Unit tests pass.

szepeviktor: Thank you.
Please do not care about `SlevomatCodingStandard.Functions.UnusedParameter.UnusedParameter`

herndlm:
Oh, of course, makes sense. Thx!

PR #141, opened 2023-02-04
Simplify returning default return type everywhere

IanDelMar:
Simplifies returning the default return type by replacing `return ParametersAcceptorSelector::selectFromArgs(...)->getReturnType();` with `return null;`.


szepeviktor:
Thank you.
That **is** improvement.

PR #143, opened 2023-02-04
Add functions to echo parameter extension

IanDelMar:
Adds the following functions to `EchoParameterDynamicFunctionReturnTypeExtension`:
* checked()
* disabled()
* menu_page_url()
* selected()
* single_month_title()
* wp_nonce_field()
* wp_original_referer_field()
* wp_readonly()
* wp_referer_field()


szepeviktor:
This is great.
How do you find these functions?

IanDelMar:
> How do you find these functions?

I checked the core files for `$echo` and `$display` arguments.

szepeviktor:
Thank you.

herndlm:
fyi @IanDelMar I just updated the extension and saw PHPStan reporting the invalid void usage usage of `wp_nonce_field()` and I thought that it found a bug, but apparently even if  `$echo` is set to true, it still returns the string. It just additionally also echoes it. E.g., see also https://developer.wordpress.org/reference/functions/wp_nonce_field/#source

not sure how other functions behave there, but this one looks like a bug.

szepeviktor:
> this one looks like a bug

👍🏻

johnbillion:
I nearly mentioned this the other day but decided not to. The original intention of the return type extension for functions with an `$echo` parameter is to basically say "this function returns void unless you set`$echo` to false" even though that's not 100% accurate because several of the functions return a value regardless.

If we correct this, several functions will always return a string even when `$echo` is true. This makes it impossible for phpstan to flag this as a problem:

```php
echo the_permalink();
```

So do we go for 100% correctness or do we go for what's most useful for developers and continue to tell a little white lie about the return value depending on `$echo`?

szepeviktor:
> we go for what's most useful for developers and continue to tell a little white lie

Okay! Lets do this one with a sentence in README.

herndlm:
my "problem" was that PHPStan started to tell me `Result of function wp_nonce_field (void) is used.` for e.g. `$nonce = wp_nonce_field('foo');` which is not true. BUT I think I get your point, I guess I should set echo to false properly, right? hmm

szepeviktor:
```
wp_nonce_field( int|string $action = -1, string $name = '_wpnonce', bool $referer = true, bool $echo = true ): string
```
WordPress is a undecipherable mess. There are only wrong answers!

`wp_nonce_field(-1, '_wpnonce', true, false)` 👉🏻 it scares me 😱

szepeviktor:
But you must say `$echo = false` or else `wp_nonce_field` echos something!

szepeviktor:
In a normal world (viktor utopia) there are different methods for each purpose.
`WordPress\Nonce::check($action)`
`WordPress\Nonce::print($action)`

IanDelMar:
> PHPStan reporting the invalid void usage usage of `wp_nonce_field()` 

This also applies to `checked()`, `disabled()`, `selected()`, `wp_readonly()`, `wp_original_referer_field()`, `wp_referer_field()`. So what we actually need is PHPStan not complaining about the return type but telling the user to properly set `$bool`.

herndlm:
I'm fine with a pragmatical approach here too. E.g. I'm gonna set the echo to false, as it should have been anyway, and at least I found this out now indirectly. Maybe we can just keep it as it is. Let's see if somebody else "complains" :)

PR #145, opened 2023-02-08
Fix return type of the_title() and single_month_title()

IanDelMar:
This PR fixes the return types for `the_title()` and `single_month_title()` in `EchoParameterDynamicFunctionReturnTypeExtension`.

I forgot to add the other types in #111 and #143.


src/EchoParameterDynamicFunctionReturnTypeExtension.php
```diff
@@ -78,13 +78,44 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
 
         return TypeCombinator::union(
-            new StringType(),
-            new VoidType()
+            self::getEchoFalseReturnType($name),
+            self::getEchoTrueReturnType($name)
         );
     }
+
+    protected static function getEchoTrueReturnType( string $name ): Type
```
szepeviktor, line 91: Why are these methods static?

src/EchoParameterDynamicFunctionReturnTypeExtension.php
```diff
@@ -78,13 +78,44 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
+            return TypeCombinator::union(
+                new VoidType(),
+                new ConstantBooleanType(false)
+            );
+        }
+
+        return new VoidType();
+    }
+
+    protected static function getEchoFalseReturnType( string $name ): Type
```
szepeviktor, line 103:
```suggestion
    protected static function getEchoFalseReturnType(string $name): Type
```

This is not WordPress :)

src/EchoParameterDynamicFunctionReturnTypeExtension.php
IanDelMar: Because they don't need an instantiation of the class

szepeviktor: Okay!

szepeviktor, line 91:
```suggestion
    protected static function getEchoTrueReturnType(string $name): Type
```

szepeviktor:
@johnbillion Do you have any comments?

szepeviktor (approved):

johnbillion (approved):
Good catches.

PR #146, opened 2023-02-09
Add return types for query strings to echo key extension

IanDelMar:
This PR addresses `EchoKeyDynamicFunctionReturnTypeExtension`. It
* adds return types for query string arguments,
* adds tests for `$args['echo']=0`, unknown values and function calls with query string,
* improves the logic of the helper methods.

### DIFF
--- a/src/EchoKeyDynamicFunctionReturnTypeExtension.php
+++ b/src/EchoKeyDynamicFunctionReturnTypeExtension.php
@@ -60,6 +60,32 @@ class EchoKeyDynamicFunctionReturnTypeExtension implements \PHPStan\Type\Dynamic
         'wp_page_menu',
     ];
 
+    /**
+     * Functions that do not accept a query string.
+     */
+    private const STRICTLY_ARRAY = [
+        'get_search_form',
+        'wp_login_form',
+    ];
+
+    /** @var \PHPStan\Reflection\FunctionReflection */
+    private $reflection;
+
+    /** @var \PHPStan\Analyser\Scope */
+    private $scope;
+
+    /** @var array<\PhpParser\Node\Arg> */
+    private $args;
+
+    /** @var string */
+    private $name;
+
+    /** @var ?int */
+    private $paramPos;
+
+    /** @var \PHPStan\Type\Type */
+    private $defaultType;
+
     public function isFunctionSupported(FunctionReflection $functionReflection): bool
     {
         return array_key_exists($functionReflection->getName(), self::SUPPORTED_FUNCTIONS);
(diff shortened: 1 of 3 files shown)


johnbillion:
I don't think `echo=false` works as you would expect. I'm pretty sure it results in string `'false'` which evaluates to true. Needs testing.

IanDelMar:
@johnbillion True!

IanDelMar:
[3e1c9d1](https://github.com/szepeviktor/phpstan-wordpress/pull/146/commits/3e1c9d1f422a5ddaf0a1679d420fe84e2abff250)
1. fixes the return types for query strings (https://github.com/szepeviktor/phpstan-wordpress/pull/146#issuecomment-1423823525),
2. removes `instanceof` (see https://github.com/szepeviktor/phpstan-wordpress/pull/147#issuecomment-1423658797),
3. reduces complexity by not checking whether a function accepts an integer as value for `echo` as WP uses loose comparison to determine whether to echo or not (`if ($args['echo]) {`)

@szepeviktor 3. requires PHPStan 1.9.4. I can either revert using the new methods or bump the minimum supported version in composer.json.

src/EchoKeyDynamicFunctionReturnTypeExtension.php
```diff
@@ -47,106 +45,109 @@ class EchoKeyDynamicFunctionReturnTypeExtension implements \PHPStan\Type\Dynamic
...
+     * Functions that do not accept a query string.
      */
-    private const STRICTLY_BOOL = [
+    private const STRICTLY_ARRAY = [
         'get_search_form',
+        'wp_login_form',
+    ];
+
+    /**
+     * Function that can return void if echo is not true/truthy.
```
szepeviktor, line 56:
```suggestion
     * Function that can return void even if `$echo` is not truthy.
```

src/EchoKeyDynamicFunctionReturnTypeExtension.php
```diff
@@ -47,106 +45,109 @@ class EchoKeyDynamicFunctionReturnTypeExtension implements \PHPStan\Type\Dynamic
...
+    private const ALWAYS_VOID = [
         'the_title_attribute',
-        'wp_list_authors',
+        'wp_dropdown_languages',
+        'wp_get_archives',
         'wp_list_comments',
-        'wp_list_pages',
-        'wp_list_users',
-        'wp_login_form',
-        'wp_page_menu',
```
szepeviktor, line 60: Where have these removed ones go?

szepeviktor:
> bump the minimum supported version in composer.json.

OK!
https://github.com/szepeviktor/phpstan-wordpress/blob/927c0f1d907d6db5757d28040ab4f0c54fdfd78e/composer.json#L16

src/EchoKeyDynamicFunctionReturnTypeExtension.php
IanDelMar: I removed the `STRICLTY_BOOL` functions because I dropped the check for int values of echo, `wp_dropdown_languages`, `wp_get_archives` and `wp_list_comments` are now part of `ALWAYS_VOID`.

szepeviktor (approved):

herndlm:
I only skimmed through it but saw first element access of array and string arrays. Jfyi in most cases the generic approach is to handle all of them and union the results then 😊

IanDelMar:
> I only skimmed through it but saw first element access of array and string arrays. Jfyi in most cases the generic approach is to handle all of them and union the results then 😊

I thought about going that route. But don't we know that there is only one array element? Generically looping through non existent elements (all but one) and then generically building a union of all the non existent elements seems pretty verbose to me. I think that in most WP cases there is no need to handle more than one element. However, if there is a clear advantage to using this general approach, I am happy to do so.

herndlm:
Don't worry, I haven't thought it through. I'm just used to that approach from phpstan internals. Here it might not add much benefits. And it could still be improved any time anyways

IanDelMar:
Would you please point me to some use cases of the generic approach? Maybe there is some benefit to it I can't imagine if I don't see it action?! Of course, doesn't have to be right now.

herndlm:
you can then support things like https://github.com/phpstan/phpstan-src/blob/3103141be78f902ff7caa207d41cc7e80d528e19/tests/PHPStan/Analyser/data/array-slice.php#L44 but this a very artificial example I guess. not sure about real world use cases, e.g. everything involving if conditions along the way like
```php
$in = $someExpression ? ['foo'] : ['bar'];
$out = array_pop($in);
// $out is 'foo'|'bar' now
```
but I guess this is also quite artifical :)

IanDelMar:
It is clearer to me now. Thank you for taking the time!

johnbillion (approved):

PR #147, opened 2023-02-09
Add dynamic return type extension for `wp_tag_cloud()`

IanDelMar:
This PR adds a dynamic function return type extension for `wp_tag_cloud()`.

I have made several updates to the extension since the initial draft:
* It now only checks for an array parameter or an empty string.
* The use of `instanceof` has been removed.
* It requires version 5.8.4+ of [wordpress-stubs](https://github.com/php-stubs/wordpress-stubs).

### DIFF
--- a/src/WpTagCloudDynamicFunctionReturnTypeExtension.php
+++ b/src/WpTagCloudDynamicFunctionReturnTypeExtension.php
@@ -57,7 +57,7 @@ class WpTagCloudDynamicFunctionReturnTypeExtension implements \PHPStan\Type\Dyna
                 : $this->getReturnType('indetermined');
         }
 
-        // Now constanst arrays are left.
+        // Now constants arrays are left.
 
         $this->setIsEchoTrue();
 
--- a/src/WpTagCloudDynamicFunctionReturnTypeExtension.php
+++ b/src/WpTagCloudDynamicFunctionReturnTypeExtension.php
@@ -1,0 +1,125 @@
+<?php
+
+/**
+ * Set return type of various functions that support an `$echo` or `$display` parameter.
+ */
+
+declare(strict_types=1);
+
+namespace SzepeViktor\PHPStan\WordPress;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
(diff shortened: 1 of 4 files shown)


IanDelMar:
I'll have a look at the failed checks tomorrow.

herndlm:
Jfyi I think we need to be careful to jot introduce too many instanceof checks, see also https://phpstan.org/blog/why-is-instanceof-type-wrong-and-getting-deprecated :)

But don't let this block you here, maybe for the next PR. We can try to refactor some away later

IanDelMar:
@herndlm Thank you! I wasn't aware of that.

src/WpTagCloudDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,125 @@
...
+
+    private static function getTypeFromString(ConstantStringType $type, string $key): ?Type
+    {
+        if (strpos($type->getValue(), "{$key}=") === false) {
+            return null;
+        }
+
+        parse_str($type->getValue(), $parsed);
+        return isset($parsed[$key])
+            ? new ConstantStringType($parsed[$key])
```
szepeviktor, line 122: > constructor expects string, array|string given

We could ignore this.

Is this PR complete?

src/WpTagCloudDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    private static function getTypeFromString(ConstantStringType $type, string $key): ?Type
+    {
+        if (strpos($type->getValue(), "{$key}=") === false) {
+            return null;
+        }
+
+        parse_str($type->getValue(), $parsed);
+        return isset($parsed[$key])
```
szepeviktor, line 121: Maybe add
```suggestion
        return isset($parsed[$key]) && is_string($parsed[$key])
```

src/WpTagCloudDynamicFunctionReturnTypeExtension.php
IanDelMar: As far as I can remember the PR is not complete.

IanDelMar:
Implemented in stubs https://github.com/szepeviktor/phpstan-wordpress/pull/147

PR #148, opened 2023-02-24
Improve `get_comment()` return type extension

IanDelMar:
This PR
* improves the dynamic return type extension for `get_comment()` by also checking the first argument and removing the null type if applicable,
* replaces the deprecated `$type instanceof ConstantStringType`,
* adds tests for the first argument being an instance of `WP_Comment` and for the no argument case.


szepeviktor:
Thank you @IanDelMar!

This looks very professional.

Issue #149, opened 2023-02-24
Remove `instanceof` from the project


szepeviktor:
❌ `instanceof`

szepeviktor:
![image](https://user-images.githubusercontent.com/952007/233822055-5e8c5df1-12c7-4878-86ca-087c27da92d8.png)

PR #150, opened 2023-02-24
Replace deprecated `$type instanceof ConstantStringType`

IanDelMar:
This PR replaces some of the deprecated `$type instanceof ConstantStringType` by
```php
if (count($type->getConstantStrings()) === 0) {}

foreach ($type->getConstantStrings() as $constantString) {}
```
as suggested in [Why Is instanceof *Type Wrong and Getting Deprecated?](https://phpstan.org/blog/why-is-instanceof-type-wrong-and-getting-deprecated).

Part of #149


szepeviktor:
Ian, this is like a gift to me.

herndlm:
👍

herndlm:
And this also gives us support for more union type scenarios as discussed recently. Thx! I also had this on my list but didn't find time yet :)

PR #153, opened 2023-02-25
Fix dynamic return type extension for `term_exists()`

IanDelMar:
This PR fixes the return type of term_exists() for an integer type first parameter with unknown value. An unknown value can be `0` for which `term_exists()` returns 0. This PR also replaces the deprecated `$type instanceof ConstantIntegerType`.

### DIFF
--- a/src/TermExistsDynamicFunctionReturnTypeExtension.php
+++ b/src/TermExistsDynamicFunctionReturnTypeExtension.php
@@ -84,12 +84,9 @@ class TermExistsDynamicFunctionReturnTypeExtension implements \PHPStan\Type\Dyna
         }
 
         foreach ($taxonomyType->getConstantStrings() as $constantString) {
-            if ($constantString->getValue() !== '') {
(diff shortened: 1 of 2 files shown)


src/TermExistsDynamicFunctionReturnTypeExtension.php
```diff
@@ -71,24 +77,20 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-                $withoutTaxonomy,
-                new NullType()
-            );
+        foreach ($taxonomyType->getConstantStrings() as $constantString) {
+            if ($constantString->getValue() !== '') {
+                $returnType[] = $withTaxonomy;
+            }
+            if ($constantString->getValue() === '') {
+                $returnType[] = $withoutTaxonomy;
+            }
```
szepeviktor, line 92:
```suggestion
            $returnType[] = $constantString->getValue() === ''
                ? $withoutTaxonomy
                : $withTaxonomy;
```

PR #155, opened 2023-02-26
Fix dynamic function return type extension for `get_sites()`

IanDelMar:
This PR
* fixes the dynamic function return type extension for `get_sites()`,
* adds correct tests,
* bumps the required minimum version of PHPStan to 1.10.0 (needed for `Type::getConstantScalarValues()`),
* removes deprecated `instanceof`.

What has been fixed?
* There is no `$args['fields'] = 'count'` but `$args['fields'] = 'ids'` and `$args['count']`. The extension now checks for both `$args['fields]` and `$args['count']`. Not properly checking for 'count' led to incorrect return types for truthy count types.
* A return type for indetermined argument type has been added.

Todo:
* Adding tests for string parameter.

### DIFF
--- a/src/GetSitesDynamicFunctionReturnTypeExtension.php
+++ b/src/GetSitesDynamicFunctionReturnTypeExtension.php
@@ -116,8 +116,8 @@ class GetSitesDynamicFunctionReturnTypeExtension implements \PHPStan\Type\Dynami
         }
 
         if (
-            in_array('ids', $fields, true) && count($fields) > 1 ||
-            !in_array('ids', $fields, true) && count($fields) > 0
+            (in_array('ids', $fields, true) && count($fields) > 1) ||
+            (!in_array('ids', $fields, true) && count($fields) > 0)
         ) {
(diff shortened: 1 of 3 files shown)


src/GetSitesDynamicFunctionReturnTypeExtension.php
```diff
@@ -36,41 +36,92 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
...
-        switch ($fields) {
-            case 'count':
-                return new IntegerType();
-            case 'ids':
-                return new ArrayType(new IntegerType(), new IntegerType());
-            default:
-                return new ArrayType(new IntegerType(), new ObjectType('WP_Site'));
+        if (
+            in_array('ids', $fields, true) && count($fields) > 1 ||
+            !in_array('ids', $fields, true) && count($fields) > 0
```
szepeviktor, line 120:
```suggestion
            (in_array('ids', $fields, true) && count($fields) > 1) ||
            (!in_array('ids', $fields, true) && count($fields) > 0)
```

szepeviktor:
Thank you Ian.

PR #156, opened 2023-02-26
Fix 'WP_Error should be referenced via a fully qualified name.'

IanDelMar:
Fixes "Exception WP_Error should be referenced via a fully qualified name." related to coding standards.
image


szepeviktor:
I think Slevomat thinks that `*Error` is an exception.

PR #157, opened 2023-02-26
Add tests for `get_sites()`

IanDelMar:
This PR adds tests for `get_sites()`
* with string parameter,
* with explicit empty parameter.


szepeviktor:
Thank you.

PR #161, opened 2023-02-26
GetSites extension: Remove duplicate values from array

IanDelMar:
This PR removes duplicate values from the `$fields` and `$count` array as the `count($this->count)` and `count($this->fields)` logic relies on unique values. This ensures that statements like `count($this->count) === 1` are `true` even if there are two or more constant arrays with `$args['count'] = true`.


szepeviktor:
Thank you.

Issue #160, opened 2023-02-26
EchoKey extension: Properly handle `getConstantStrings()` and `getConstantArrays()`


IanDelMar:
> I only skimmed through it but saw first element access of array and string arrays. Jfyi in most cases the generic approach is to handle all of them and union the results then 😊

_Originally posted by @herndlm in https://github.com/szepeviktor/phpstan-wordpress/issues/146#issuecomment-1426219522_

Loop through all elements in the list of constant strings/array:

```php
if (count($type->getConstantStrings()) === 0) {
    return []; // no constant strings
}

foreach ($type->getConstantStrings() as $constantString) {
    // do something with each value
}
```

PR #162, opened 2023-02-26
Improve echo type handling in EchoKey extension

herndlm:
Closes https://github.com/szepeviktor/phpstan-wordpress/issues/160

@IanDelMar check out the first one, we don't even have to care if it's a constant array or not :)

the second one feels a bit like an overkill, but it is more correct I guess..


IanDelMar:
The first one is much better than what I came up with. Don't care about the second one; support for query strings will be dropped anyways.

Issue #154, opened 2023-02-25
Incorrect return type for `esc_sql()`, `wp_slash()` and `wp_unslash()`


IanDelMar:
`StringOrArrayDynamicFunctionReturnTypeExtension` narrows the return type to strings. While this is true for `esc_sql()`, `wp_slash()` and `wp_unslash()` do not change the type of the values. 
```php
wp_slash([true]); // gives [true]
wp_unslash(1); // gives 1
esc_sql([true]); // gives ['1']
esc_sql(1); // gives '1'
```

For `Array[]` also the return type for `esc_sql()` is incorrect.
```php
esc_sql([[1]]); // gives array<int,array<int,string>>
esc_sql([['key' => 1]]); // gives array<int,array<string,string>>
esc_sql(['key' => [1]]); // gives array<string,array<int,string>>
esc_sql(['key1' => ['key2' => 1]]); // gives array<string,array<string,string>>
// the extension gives either array<int,string> or array<string,string>
```

Currently the extension also lacks a maybe array case, ie `string|array<int|string,mixed>`, and a maybe string array key case, ie `array<int|string,mixed>`.

IanDelMar:
Prior to WP 5.5.0 non-string values in `wp_slash()` and `wp_unslash()` were not left untouched. From v5.9.4 of [php-stubs/wordpress-stubs](https://github.com/php-stubs/wordpress-stub) `wp_slash()` and `wp_unslash()` are covered in the WP stubs.

@szepeviktor Since the return type extension takes precedence over the stubs files and given that v5.9+ accounts for 71.4% of installed versions I suggest to drop `wp_slash()` and `wp_unslash()` all together. For the majority of WP installs correct return types would be given via the stubs file. 83.1% of installs see incorrect return types, while only 16.9% of install benefit from the outdated extension. (Assuming that the distribution of installed WP versions is equal among PHPStan users and all users.) What do you think?

szepeviktor:
I go with your decision.
https://developer.wordpress.org/reference/functions/wp_slash/
I use only the good parts of WordPress, thus no arrays into wp_slash.

PR #163, opened 2023-02-26
Get rid of instanceof in StringOrArray extension

herndlm:


IanDelMar:
I'm not quite sure that this is sufficient. While `$keyType instanceof StringType` was a boolean `$keyType->isString()` is a trinary logic. We should cover `$keyType->isString()->maybe()` as well, don't we?

herndlm:
Not sure tbh, this does not change behaviour at least to before, except e.g. supporting unions of strings too. Maybes like mixed were also not considered before. It could still be improved any time I guess

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
```diff
@@ -35,7 +35,7 @@ public function getTypeFromFunctionCall(FunctionReflection $functionReflection,
         $dataArgType = $scope->getType($dataArg);
         if ($dataArgType->isArray()->yes()) {
             $keyType = $dataArgType->getIterableKeyType();
-            if ($keyType instanceof StringType) {
+            if ($keyType->isString()->yes()) {
                 return new ArrayType(new StringType(), new StringType());
             }
             return new ArrayType(new IntegerType(), new StringType());
```
IanDelMar, line 41:
```suggestion
            $returnType = [new ArrayType(new IntegerType(), new StringType())];
            if ($keyType->isString()->maybe()) {
                $returnType[] = new ArrayType(new StringType(), new StringType());
            }
            return TypeCombinator::union(...$returnType);
```

IanDelMar (changes requested):

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
IanDelMar, line 41:
```suggestion
            if ($keyType->isString()->no()) {
                return new ArrayType(new IntegerType(), new StringType());
            }
            return TypeCombinator::union(
                new ArrayType(new StringType(), new StringType()),
                new ArrayType(new IntegerType(), new StringType())
            );
```

IanDelMar (changes requested):

IanDelMar:
I agree, it is as good as before, it just makes it more obvious that the unknown key type case was not covered.

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
herndlm: I see what you're going for I think. I don't really understand though what this extension is even doing :) I was just replacing stuff without changing much behaviour.
Why does it not just pass through the array but recreate it for example. Unfortunately not  on my computer anymore to dig deeper..

IanDelMar: It sets the value type to string no matter what. However, this is not what the extension was intended to I think. See #154.

herndlm: I'll check it later and try to add tests too it not existing yet. But does it make sense to you to directly pass through the key type? It should noy be modified, right?

PR #167, opened 2023-02-27
Improve/fix type StringOrArray extension

herndlm:
Closes https://github.com/szepeviktor/phpstan-wordpress/issues/154

- Gets rid of deprecated `instanceof`
- Fixes return types for nested arrays
- Adds more type precision by not dropping array key types, supporting constant arrays and returning `''` in case of an invalid type input
- Adds support for union/intersection types

### DIFF
--- a/src/StringOrArrayDynamicFunctionReturnTypeExtension.php
+++ b/src/StringOrArrayDynamicFunctionReturnTypeExtension.php
@@ -11,10 +11,16 @@ namespace SzepeViktor\PHPStan\WordPress;
 use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\FunctionReflection;
-use PHPStan\Type\StringType;
 use PHPStan\Type\ArrayType;
-use PHPStan\Type\IntegerType;
+use PHPStan\Type\Constant\ConstantArrayType;
+use PHPStan\Type\Constant\ConstantArrayTypeBuilder;
+use PHPStan\Type\Constant\ConstantStringType;
+use PHPStan\Type\IntersectionType;
+use PHPStan\Type\StringType;
 use PHPStan\Type\Type;
+use PHPStan\Type\TypeCombinator;
+use PHPStan\Type\TypeTraverser;
+use PHPStan\Type\UnionType;
 
 class StringOrArrayDynamicFunctionReturnTypeExtension implements \PHPStan\Type\DynamicFunctionReturnTypeExtension
 {
@@ -29,21 +35,63 @@ class StringOrArrayDynamicFunctionReturnTypeExtension implements \PHPStan\Type\D
     // phpcs:ignore SlevomatCodingStandard.Functions.UnusedParameter
     public function getTypeFromFunctionCall(FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope): ?Type
     {
-        $args = $functionCall->getArgs();
-        if (count($args) === 0) {
+        if (count($functionCall->getArgs()) === 0) {
             return null;
         }
 
-        $dataArgType = $scope->getType($args[0]->value);
-        if ($dataArgType->isArray()->yes()) {
-            $keyType = $dataArgType->getIterableKeyType();
-            if ($keyType instanceof StringType) {
-                return new ArrayType(new StringType(), new StringType());
-            }
+        return TypeTraverser::map(
+            $scope->getType($functionCall->getArgs()[0]->value),
+            function (Type $type, callable $traverse): Type {
+                if ($type instanceof UnionType || $type instanceof IntersectionType) {
+                    // phpcs:ignore NeutronStandard.Functions.VariableFunctions.VariableFunction
+                    return $traverse($type);
+                }
 
-            return new ArrayType(new IntegerType(), new StringType());
+                return $this->getType($type);
+            }
+        );
+    }
+
+    private function getType(Type $type): Type
+    {
+        if ($type->isScalar()->yes()) {
+            return new StringType();
         }
 
-        return new StringType();
+        if (!$type->isArray()->yes()) {
+            return new ConstantStringType('');
+        }
+
+        if (count($type->getConstantArrays()) > 0) {
+            return TypeCombinator::union(
+                ...array_map(
+                    function (ConstantArrayType $constantArray): Type {
+                        $builder = ConstantArrayTypeBuilder::createEmpty();
+                        foreach ($constantArray->getKeyTypes() as $i => $keyType) {
+                            $builder->setOffsetValueType(
+                                $keyType,
+                                $this->getType($constantArray->getValueTypes()[$i]),
+                                $constantArray->isOptionalKey($i)
+                            );
+                        }
+
+                        return $builder->getArray();
+                    },
+                    $type->getConstantArrays()
+                )
+            );
+        }
+
+        return TypeCombinator::union(
+            ...array_map(
+                function (ArrayType $arrayType): Type {
+                    return $arrayType->setOffsetValueType(
+                        $arrayType->getIterableKeyType(),
+                        $this->getType($arrayType->getIterableValueType())
+                    );
+                },
+                $type->getArrays()
+            )
+        );
     }
 }
--- a/tests/data/esc_sql.php
+++ b/tests/data/esc_sql.php
@@ -9,13 +9,35 @@ use stdClass;
 use function PHPStan\Testing\assertType;
 
 // Indexed array as parameter
-assertType('array<int, string>', esc_sql(['someValue', 'toEscape']));
+assertType("array{string, string}", esc_sql(['someValue', 'toEscape']));
 
 // Associative array as parameter
-assertType('array<string, string>', esc_sql(['someValue' => 'toEscape']));
+assertType("array{someValue: string}", esc_sql(['someValue' => 'toEscape']));
 
 // String as parameter
 assertType('string', esc_sql('someValueToEscape'));
 
-// Wrong type provided (esc_sql() returns an empty string in that case)
-assertType('string', esc_sql(new stdClass()));
+// Wrong type provided
+assertType("''", esc_sql(new stdClass()));
+assertType("''", esc_sql(null));
+assertType("array{''}", esc_sql([null]));
+
+assertType('array{string}', esc_sql([true]));
+assertType('string', esc_sql(1));
+assertType('array{array{string}}', esc_sql([[1]]));
+assertType('array{array{key: string}}', esc_sql([['key' => 1]]));
+assertType('array{key: array{string}}', esc_sql(['key' => [1]]));
+assertType('array{key1: array{key2: string}}', esc_sql(['key1' => ['key2' => 1]]));
+
+/** @var array{foo?: 'something'}|array{bar: 'something else'} $union1 */
+$union1 = null;
+assertType("array{bar: string}|array{foo?: string}", esc_sql($union1));
+
+/** @var 'foo'|array{foo?: 'bar'} $union2 */
+$union2 = null;
+assertType("array{foo?: string}|string", esc_sql($union2));
+
+/** @var array{foo?: 'bar'}|null $union3 */
+$union3 = null;
+assertType("''|array{foo?: string}", esc_sql($union3));
+


src/StringOrArrayDynamicFunctionReturnTypeExtension.php
```diff
@@ -29,21 +35,63 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+        if ($type->isScalar()->yes()) {
+            return new StringType();
+        }
+
+        if (!$type->isArray()->yes()) {
+            return new ConstantStringType('');
+        }
+
+        if (count($type->getConstantArrays()) > 0) {
+            return TypeCombinator::union(
```
herndlm, line 66: those wouldn't even be needed anymore I guess because this function is wrapped in a TypeTraverser already that deals with union types. the alternative would be to explicitly handle only the first constant array and array below instead of all of them. not sure what is simpler / reads better, in the end it might not matter much..

tests/data/esc_sql.php
```diff
@@ -9,13 +9,35 @@
 use function PHPStan\Testing\assertType;
 
 // Indexed array as parameter
-assertType('array<int, string>', esc_sql(['someValue', 'toEscape']));
+assertType("array{string, string}", esc_sql(['someValue', 'toEscape']));
```
szepeviktor, line 12:
```suggestion
assertType('array{string, string}', esc_sql(['someValue', 'toEscape']));
```

tests/data/esc_sql.php
```diff
@@ -9,13 +9,35 @@
...
 
 // Associative array as parameter
-assertType('array<string, string>', esc_sql(['someValue' => 'toEscape']));
+assertType("array{someValue: string}", esc_sql(['someValue' => 'toEscape']));
```
szepeviktor, line 15:
```suggestion
assertType('array{someValue: string}', esc_sql(['someValue' => 'toEscape']));
```

tests/data/esc_sql.php
```diff
@@ -9,13 +9,35 @@
...
+assertType('array{string}', esc_sql([true]));
+assertType('string', esc_sql(1));
+assertType('array{array{string}}', esc_sql([[1]]));
+assertType('array{array{key: string}}', esc_sql([['key' => 1]]));
+assertType('array{key: array{string}}', esc_sql(['key' => [1]]));
+assertType('array{key1: array{key2: string}}', esc_sql(['key1' => ['key2' => 1]]));
+
+/** @var array{foo?: 'something'}|array{bar: 'something else'} $union1 */
+$union1 = null;
+assertType("array{bar: string}|array{foo?: string}", esc_sql($union1));
```
szepeviktor, line 34:
```suggestion
assertType('array{bar: string}|array{foo?: string}', esc_sql($union1));
```

tests/data/esc_sql.php
```diff
@@ -9,13 +9,35 @@
...
+
+/** @var 'foo'|array{foo?: 'bar'} $union2 */
+$union2 = null;
+assertType("array{foo?: string}|string", esc_sql($union2));
```
szepeviktor, line 38:
```suggestion
assertType('array{foo?: string}|string', esc_sql($union2));
```

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
```diff
@@ -29,21 +35,63 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+                    },
+                    $type->getConstantArrays()
+                )
+            );
         }
 
-        return new StringType();
+        return TypeCombinator::union(
+            ...array_map(
+                function (ArrayType $arrayType): Type {
```
szepeviktor, line 87:
```suggestion
                static function (ArrayType $arrayType): Type {
```

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
```diff
@@ -29,21 +35,63 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+        }
+
+        if (!$type->isArray()->yes()) {
+            return new ConstantStringType('');
+        }
+
+        if (count($type->getConstantArrays()) > 0) {
+            return TypeCombinator::union(
+                ...array_map(
+                    function (ConstantArrayType $constantArray): Type {
```
szepeviktor, line 68:
```suggestion
                    static function (ConstantArrayType $constantArray): Type {
```

tests/data/esc_sql.php
herndlm: right.. I had the wrong type there initially and forgot to remove them. removed all in a dedicated commit

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
```diff
@@ -29,21 +35,63 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
         }
 
-        $dataArgType = $scope->getType($args[0]->value);
-        if ($dataArgType->isArray()->yes()) {
-            $keyType = $dataArgType->getIterableKeyType();
-            if ($keyType instanceof StringType) {
-                return new ArrayType(new StringType(), new StringType());
+        return TypeTraverser::map(
+            $scope->getType($functionCall->getArgs()[0]->value),
+            function (Type $type, callable $traverse): Type {
```
szepeviktor, line 44:
```suggestion
            static function (Type $type, callable $traverse): Type {
```

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
herndlm: this is not possible because it uses recursion in the callable

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
szepeviktor: Okay.

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
herndlm: ditto, recursion via `$this` :)

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
herndlm: also not possible, uses `$this->getType()` :)

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
szepeviktor: Oh  noooooooooooooooooooooooooooooooooooooooooooo!

szepeviktor:
Does PHP allow `$this` in a closure? I would not allow it!

herndlm:
in theory 

> Does PHP allow `$this` in a closure? I would not allow it!

sure, there's no reason against it. without such code (and also e.g. with the weird Neutron CS rule that forbids to have a callable assigned to a var) PHPStan wouldn't be able to efficiently parse and handle the AST to figure out types

herndlm:
@IanDelMar let me/us know please what you think of this. I did briefly confirm the test cases with a real WordPress, but I don't know the internals of the escaping function well enough. Maybe @johnbillion can quickly check too :)

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
IanDelMar: @herndlm I also think that `if (count($type->getConstantArrays()) > 0) {...}` is no longer needed.

tests/data/esc_sql.php
```diff
@@ -9,13 +9,35 @@
...
 assertType('string', esc_sql('someValueToEscape'));
 
-// Wrong type provided (esc_sql() returns an empty string in that case)
-assertType('string', esc_sql(new stdClass()));
+// Wrong type provided
+assertType("''", esc_sql(new stdClass()));
+assertType("''", esc_sql(null));
+assertType("array{''}", esc_sql([null]));
+
+assertType('array{string}', esc_sql([true]));
```
IanDelMar, line 25:
```suggestion
assertType('string', esc_sql(true));
assertType('array{string}', esc_sql([true]));
```

IanDelMar:
@herndlm Looks good to me!

szepeviktor:
Is it ready to merge?

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
```diff
@@ -29,21 +35,63 @@ public function isFunctionSupported(FunctionReflection $functionReflection): boo
...
+        );
+    }
+
+    private function getType(Type $type): Type
+    {
+        if ($type->isScalar()->yes()) {
+            return new StringType();
+        }
+
+        if (!$type->isArray()->yes()) {
```
IanDelMar, line 61:
```suggestion
        if ($type->isArray()->no()) {
```

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
IanDelMar: Had a closer look. What about this?

```php
        $maybeEmptyArray = [];
        if (count($type->getConstantArrays()) > 0) {
            $constantArray = $type->getConstantArrays()[0];
            if (
                $constantArray->isOptionalKey(0) &&
                $constantArray->getArraySize()->equals(IntegerRangeType::fromInterval(0, 1))
            ) {
                $maybeEmptyArray[] = $constantArray;
            }
        }

        return TypeCombinator::union(
            ...array_merge(
                $maybeEmptyArray,
                array_map(
                    function (ArrayType $arrayType): Type {
                        return $arrayType->setOffsetValueType(
                            $arrayType->getIterableKeyType(),
                            $this->getType($arrayType->getIterableValueType())
                        );
                    },
                    $type->getArrays()
                )
            )
        );
```

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
szepeviktor: If it is a maybe type !yes means maybe or no,
!no means maybe or yes.

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
IanDelMar: No, this fails for
```php
/** @var 'foo'|array{foo?: true, bar: int} $union */
$union = null;
assertType('array{foo?: string, bar: string}|string', esc_sql($union));
```
with
```
-'array{foo?: string, bar: string}|string'
+'array{foo?: string|true, bar: int|string}|string'
```

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
herndlm: Both are not a 100% correct I suppose, the maybe case might need another extra handling instead of just returning the input below. I'll check it later.

herndlm: dealing with the maybe here is highly annoying, e.g. your proposed change would lead to a `*NEVER*` (mine to a `''`) when a maybe array like `mixed` or `iterable` is passed. both are wrong. and in both cases PHPStan should be complaining because esc_sql does not support those types via PHPDoc. Natively it does of course, I know..

And I can't even say what we would return with a mixed, I guess we could just bail out early and not handle it or smth like that

herndlm: adapted, what do you think? should at least be simpler and less wrong now :)

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
herndlm: got rid of the array_map and union

src/StringOrArrayDynamicFunctionReturnTypeExtension.php
szepeviktor: @IanDelMar we listen to you

IanDelMar: > And I can't even say what we would return with a mixed, I guess we could just bail out early and not handle and return the default type or smth like that

Yep, I would go for returning null if everything else does not apply. That's what we're doing in other extensions as well, no? Dealing with what we want to deal with and return the default otherwise. It's still an improvement and doesn't do worse in any case.

herndlm: Thought the same, but can't return null here because this is wrapped with the TypeTraverser:map() which doesn't allow returning null AFAIK

herndlm: BTW just curious, your profile says Vienna but you seem yo be active very late until into the night. Ate you still in Vienna? I'm from Austria and lived there too, missing it a bit from time to time..

IanDelMar: I still live in Vienna. My activity time mainly depends on my day job and after work napping :-D. Vienna is a nice place to live. I would miss it too at least in late spring and summer.

herndlm: And good old boss Viktor is in Budapest, the twin cities, right? :)
Anyway, shall we merge? :P

szepeviktor: I *am* old.

PR #169, opened 2023-03-02
Add Support my work section

szepeviktor:
👀
https://github.com/szepeviktor/phpstan-wordpress/blob/szepeviktor-patch-1/README.md#support-my-work
@adamlaki There seems to be a color difference and this heart does not beat :(

What if we add the same ids/classes -> then GitHub's CSS will work on our svg too?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,14 @@ includes:
     - vendor/szepeviktor/phpstan-wordpress/extension.neon
 ```
 
+### Support my work
+
+Please consider supporting my work.
+
+[![Sponsor](https://github.com/szepeviktor/.github/raw/master/.github/assets/github-like-sponsor-button.svg)](https://github.com/sponsors/szepeviktor)
+
+Thank you!
+
 ### Configuration
 
 Needs no extra configuration. :smiley: Simply configure PHPStan - for example - this way.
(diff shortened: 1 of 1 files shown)


adamlaki:
Viktor,

you asked this via e-mail; why didn't you continue the discussion over there?

Please, don't give half the information. If a color does not match, what color is it?

What kind of animation do you want, just on the hover state (like the button)?

szepeviktor:
> Please, don't give half the information. If a color does not match, what color is it?

I'm sorry.
This PR fixes the color. https://github.com/szepeviktor/.github/pull/1
Could you approve it?

> What kind of animation do you want, just on the hover state (like the button)?

Yes, hover-only animation.
Is that possible in SVG?
This button is used in README-s.

adamlaki:
I've approved it, looks good.

Yes, in theory, it is possible. I will make the addition soon.

adamlaki:
I don't understand where the color diff is coming from; it is correct in the file that I sent you.

szepeviktor:
I've added some SVG features and optimization.

```xml
    <rect width="92" height="28" ...
    <path fill="#BF3989" ...
    <text ...
```

PR #172, opened 2023-03-28
Add some more dynamic constants

johnbillion:
This list is potentially endless, but here are two more constants that I've had to add to the `dynamicConstantNames` list in my own plugins.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -120,6 +120,8 @@ parameters:
         - WP_DEBUG_LOG
         - EMPTY_TRASH_DAYS
         - WP_CLI
+        - COOKIE_DOMAIN
+        - SAVEQUERIES
     earlyTerminatingFunctionCalls:
         - wp_die
         - wp_send_json


szepeviktor:
> This list is potentially endless

You are right.
We have this rotten tomato for globals: https://github.com/php-stubs/wordpress-globals

Constants in this project contain only "popular" ones.

szepeviktor:
For you John, I would merge this one if you need it badly!

szepeviktor:
I see! You need both `dynamicConstantNames` **and** stubs.

szepeviktor:
```php
$domain = COOKIE_DOMAIN ?: '';
```

Please help me. When is `COOKIE_DOMAIN` falsy?

johnbillion:
Its default value is `false` if it's not defined anywhere else. https://github.com/WordPress/wordpress-develop/blob/dbbc9bcf98787fd77fd920df84cc9bbc69974b0a/src/wp-includes/default-constants.php#L317

szepeviktor:
Thank you.

PR #173, opened 2023-04-07
Fix earlyTerminatingMethodCalls syntax

mundschenk-at:
Not sure if that was a change with recent PHPStan versions or always incorrect, but these only work when I follow the syntax example from https://phpstan.org/writing-php-code/solving-undefined-variables

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -137,16 +137,16 @@ parameters:
         - wp_ajax_heartbeat
         - wp_ajax_nopriv_heartbeat
     earlyTerminatingMethodCalls:
-        \Custom_Background:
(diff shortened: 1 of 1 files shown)


szepeviktor:
A surprise PR! Thank you @mundschenk-at!

Could you add a test that fails without this PR?

szepeviktor:
@dingo-d Started it 3 years ago.

dingo-d:
> @dingo-d started it 3 years ago.

I did? 😅

szepeviktor:
Thank you @mundschenk-at

PR #177, opened 2023-04-13
Add dynamic return type extension for `get_post_types`

lipemat:
Introduce GetPostTypesDynamicFunctionReturnTypeExtension service

The `get_post_types` function return either and array of `string` or an array of `WP_Post_Type` depending on the passed arguments.

https://developer.wordpress.org/reference/functions/get_post_types/#return

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -107,6 +107,10 @@ services:
         class: SzepeViktor\PHPStan\WordPress\AssertNotWpErrorTypeSpecifyingExtension
         tags:
             - phpstan.typeSpecifier.methodTypeSpecifyingExtension
+    -
+        class: SzepeViktor\PHPStan\WordPress\GetPostTypesDynamicFunctionReturnTypeExtension
+        tags:
+            - phpstan.broker.dynamicFunctionReturnTypeExtension
 rules:
     - SzepeViktor\PHPStan\WordPress\HookCallbackRule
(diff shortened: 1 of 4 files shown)


szepeviktor:
Could it be done with the new conditional type in WP stubs?

herndlm:
> Could it be done with the new conditional type in WP stubs? `@phpstan-assert-if...` https://phpstan.org/writing-php-code/narrowing-types#custom-type-checking-functions-and-methods

good point, looks like this would work: https://phpstan.org/r/6e188841-6faf-4de8-9c0f-57ebabc3e626

lipemat:
> can you add a new file with assertion tests, please? an example would be e.g. #144

Tests are now included should you decide to use this code. :)

szepeviktor:
Thank you for your work.
I am ver sorry. That conditional thing, Martin wrote, makes PHPStan work alone. That should be added to WP stubs.

IanDelMar:
The conditional thing then must work for `get_taxonomies()` as well.

szepeviktor:
@lipemat Conditional return types arrived in https://github.com/php-stubs/wordpress-stubs/pull/73

Thank you for your PR!!

szepeviktor:
I will use your tests.
`tests/data/get_post_types.php`

PR #178, opened 2023-04-13
Fix CS

herndlm:
not making the pipeline green, but fixes the cs step at least..


herndlm:
there are 3 extension files with phpstan errors left, they are tricky to refactor, but I'll try looking into it

szepeviktor:
VERY nice! We support a discouraged constant.

herndlm:
Even my name is written properly in katakana, not bad! But this is really nothing special 😊 there's more work to do...

PR #183, opened 2023-04-21
Remove deprecated `instanceof` from `wp_parse_url` extension

IanDelMar:
Part of #149


szepeviktor:
I'm so glad there will be one less.

PR #184, opened 2023-04-22
Remove deprecated `instanceof`

IanDelMar:
What I was wondering when updating this extension. I can have a non constant array that has a key `'fields'`.
```php
// Non constant array with fields first
$array = [
    'fields' => 'ids',
    $_GET['foo'] => $_GET['bar'],
];
assertType('array<int, int|WP_Post>', get_posts($array));

// Non constant array with fields last
$array = [
    $_GET['foo'] => $_GET['bar'],
    'fields' => 'ids',
];
assertType('array<int, int>', get_posts($array));
```

`get_post($array)` with the fields first case returns `array<int, int|WP_Post>` because `$_GET['foo']` may be `'fields'` and overwrite `'ids'`. `get_post($array)` with the fields last case returns  `array<int, int>` as `'fields'` is not affected by `$_GET['foo']` and hence will  be `'ids'`. With limited PHPStan knowledge, I was not able to find a way to check the order of `$_GET['foo']` and `'fields'` or more specifically to check whether there is a non constant key after the fields key.

Is there a simple way to check the order of the keys with PHPStan?

Part of #149

### DIFF
--- a/tests/data/get_posts.php
+++ b/tests/data/get_posts.php
@@ -55,3 +55,12 @@ assertType('array<int, int|WP_Post>', get_posts(['fields' => $union]));
 
 $union = $_GET['foo'] ? (string)$_GET['string'] : 'id=>parent';
 assertType('array<int, int|WP_Post>', get_posts(['fields' => $union]));
+
+$union = $_GET['foo'] ? (string)$_GET['string'] : 'fields';
+assertType('array<int, int|WP_Post>', get_posts([$union => '']));
+
+$union = $_GET['foo'] ? (string)$_GET['string'] : 'fields';
+assertType('array<int, int|WP_Post>', get_posts([$union => 'ids']));
+
+$union = $_GET['foo'] ? (string)$_GET['string'] : 'fields';
+assertType('array<int, int|WP_Post>', get_posts([$union => 'id=>parent']));
--- a/tests/data/get_posts.php
+++ b/tests/data/get_posts.php
@@ -46,3 +46,12 @@ assertType('array<int, int|WP_Post>', get_posts($union));
 
 $union = $_GET['foo'] ? (array)$_GET['array'] : ['fields' => 'id=>parent'];
 assertType('array<int, int|WP_Post>', get_posts($union));
+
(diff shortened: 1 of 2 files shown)


tests/data/get_posts.php
```diff
@@ -7,6 +7,42 @@
...
 
 assertType('array<int, WP_Post>', get_posts());
-assertType('array<int, WP_Post>', get_posts(['fields' => 'all']));
+assertType('array<int, WP_Post>', get_posts(['key' => 'value']));
+assertType('array<int, WP_Post>', get_posts(['fields' => '']));
 assertType('array<int, int>', get_posts(['fields' => 'ids']));
 assertType('array<int, int>', get_posts(['fields' => 'id=>parent']));
+assertType('array<int, WP_Post>', get_posts(['fields' => 'Hello']));
+
+// Non constant array
```
szepeviktor, line 16:
```suggestion
// Nonconstant array
```

Please see https://www.merriam-webster.com/dictionary/nonconstant

szepeviktor:
We could open a ticket at @phpstan and talk to @ondrejmirtes about a function changing its return type based on its parameter `'fields' => 'ids'` which is 😢 an array.
This is so anti-PHPStan, I am so ashamed.

Let's support only usage where there is only 1 `fields` element (anywhere) in the array.

IanDelMar:
I'd like to add more tests. Please wait with merging.

IanDelMar:
> We could open a ticket at @phpstan and talk to @ondrejmirtes about a function changing its return type based on its parameter `'fields' => 'ids'` which is 😢 an array.

I also thought we need something like that.

IanDelMar:
I can't think of any other test cases to add. Do you?

szepeviktor:
We already have much more tests than possible problems.

Thank you.

herndlm:
How exactly is the type looking in your more dynamic example? Is the key just string then? Maybe it helps if you do checks for the key of the whole array? string|fields would be generalised to string and then you can handle this case differently maybe? Or I misunderstood the problem :)

szepeviktor:
We are digging too deep.
Let's fix future problems in the next PR!

PR #185, opened 2023-04-23
Remove deprecated `instanceof`

IanDelMar:
I also removed dealing with the `$count` argument which was deprecated in WP 4.5. In WP 4.5 also the position of the `$args` argument changed. Given  we use `'get_terms' => 0,` we're not supporting WP versions lower than 4.5. Hence `$count` had no effect.

Travis should be going green again soon. This was the last extension with deprecation notices.

Part of #149


szepeviktor:
> Travis should be going green again soon

You did it Ian!

image

herndlm:
Great work!

PR #190, opened 2023-06-16
Fix _get_list_table extension

IanDelMar:
Fixes the dynamic return type extension for `_get_list_table()`.

According to the [docs](https://developer.wordpress.org/reference/functions/_get_list_table/), `_get_list_table()` returns a (known or unknown) child class of WP_List_Table or WP_List_Table class itself if the passed argument is a WP_List_Table (child) class name, and `false` otherwise.

As the class name is filterable any WP_List_Table child class maybe returned... We can either assume that in most cases this filtered is not used and return a `WP_Posts_List_Table` when the argument is `WP_Posts_List_Table` and risk that this is incorrect (the previous functionality and the current approach) or we can simply return a `WP_List_Table` which will be correct but less precise.


szepeviktor:
Thank you!

PR #192, opened 2023-06-17
Remove unnecessary tests for mysql2date extension

IanDelMar:
Drops tests for unsupported types as PHPStan warns about the incorrect parameter types.


szepeviktor:
Thank you.

PR #201, opened 2023-10-14
Add dynamic return type extension for `wp_die()`

### LINKED ISSUE #200: `wp_die()` not exclusively an early terminating function
`wp_die()` is treated as if it terminated script execution early:
https://github.com/szepeviktor/phpstan-wordpress/blob/76f783564edc0bd8f4f3a2e44f6c09a1092702d5/extension.neon#L126-L127

But for `wp_die('', '', ['exit' => false])` the script execution will not be terminated in any of WP's wp_die handlers.

```php
if (somethingIsTrue()) {
	$foo = true;
} elseif (orSomethingElseIsTrue()) {
	$foo = false;
} else {
	wp_die('', '', ['exit' => false])
}

doFoo($foo); // won't catch possibly undefined variable $foo
```

IanDelMar:
Fixes #200

### DIFF
--- a/tests/data/wp_die.php
+++ b/tests/data/wp_die.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SzepeViktor\PHPStan\WordPress\Tests;
+
+use function PHPStan\Testing\assertType;
+
+/** @var array $array */
+$array = null;
+
+$true = ['exit' => true];
+$false = ['exit' => false];
+
+assertType('*NEVER*', wp_die('', ''));
+assertType('*NEVER*', wp_die('', '', $true));
+assertType('void', wp_die('', '', $false));
+assertType('void', wp_die('', '', $array));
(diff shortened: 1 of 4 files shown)


tests/data/wp_die.php
```diff
@@ -0,0 +1,18 @@
...
+
+/** @var array $array */
+$array = null;
+
+$true = ['exit' => true];
+$false = ['exit' => false];
+
+assertType('*NEVER*', wp_die('', ''));
+assertType('*NEVER*', wp_die('', '', $true));
+assertType('void', wp_die('', '', $false));
```
szepeviktor, line 17: Could we simplify this?

```suggestion
assertType('*NEVER*', wp_die('', ''));
assertType('*NEVER*', wp_die('', '', ['exit' => true]));
assertType('void', wp_die('', '', ['exit' => false]));
```

szepeviktor:
The new stubs cause comment and taxonomy errors?

szepeviktor:
Thank you.

Hereby I unban you from the Internet 🙃

PR #206, opened 2023-11-01
Add test for WpThemeMagicPropertiesClassReflectionExtension

IanDelMar:


szepeviktor:
Thank you.

tests/WpThemeMagicPropertiesClassReflectionExtensionTest.php
```diff
@@ -0,0 +1,49 @@
...
+
+namespace SzepeViktor\PHPStan\WordPress\Tests;
+
+use SzepeViktor\PHPStan\WordPress\WpThemeMagicPropertiesClassReflectionExtension;
+
+class WpThemeMagicPropertiesClassReflectionExtensionTest extends \PHPStan\Testing\PHPStanTestCase
+{
+    protected function setUp(): void
+    {
+        $this->broker = $this->createBroker();
```
szepeviktor, line 13: Where does broker come from?

szepeviktor (changes requested):

tests/WpThemeMagicPropertiesClassReflectionExtensionTest.php
szepeviktor: From Ian!

tests/WpThemeMagicPropertiesClassReflectionExtensionTest.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace SzepeViktor\PHPStan\WordPress\Tests;
+
+use SzepeViktor\PHPStan\WordPress\WpThemeMagicPropertiesClassReflectionExtension;
+
+class WpThemeMagicPropertiesClassReflectionExtensionTest extends PHPStanTestCase
```
szepeviktor, line 9:
```suggestion
class WpThemeMagicPropertiesClassReflectionExtensionTest extends \PHPStan\Testing\PHPStanTestCase
```

szepeviktor:
🎵 Payment for the PR: Shamisen plunking 🇯🇵 https://www.youtube.com/watch?v=RcqO0zkyQRo

IanDelMar:
Very generous! :dancers:

szepeviktor:
... we could have a non-existent property - but do not worry about it

IanDelMar:
Do you mean that `WP_Theme` does not necessarily have all of these properties? I var_dumped all  properties of a theme where I removes all info from style.css except the name of the theme. All properties get populated at least with an empty string or an empty array.

IanDelMar:
But now I'm wondering why we don't add them via the stubs' functionMap.

szepeviktor:
I've meant to add a line like
```
['WP_Theme', 'gluegun', false],
```

IanDelMar:
https://github.com/szepeviktor/phpstan-wordpress/blob/2e08b58b0060b7f878244a36e5481ff51815bd43/tests/WpThemeMagicPropertiesClassReflectionExtensionTest.php#L25

szepeviktor:
You see?? The PR is longer than 1 line so viktor is 😵 confuzed!

PR #208, opened 2023-11-02
Fix dynamic call to static method

IanDelMar:
The `\PHPStan\Testing\TypeInferenceTestCase::gatherAssertTypes()` method is called statically. See the implementation in the PHPStan source code here: [PHPStan TypeInferenceTestCase.php line 118-120](https://github.com/phpstan/phpstan-src/blob/6bdc25a944690839507a8ca96d33ec7c51660f3f/src/Testing/TypeInferenceTestCase.php#L118-L120).

We have updated the minimum required version from 1.10.0 to 1.10.31. In version 1.10.0, it was still called dynamically, as can be seen here: [PHPStan TypeInferenceTestCase.php line 81-83](https://github.com/phpstan/phpstan-src/blob/2ca979344d7750dee065ffb6c43af6236f398241/src/Testing/TypeInferenceTestCase.php#L81-L83)."


szepeviktor:
All right!

PR #211, opened 2024-02-23
Introduce dynamic return type for the `get_approved_comments` function

lipemat:
`get_approved_comments` changes the return type based on value within the passed $args parameter.

- Return 'WP_Comment[]' by default.
- Return `int[]` if `$fields = 'ids'`.
- Return `int` if `$count = true`.

https://developer.wordpress.org/reference/functions/get_approved_comments/#return

### DIFF
--- a/src/GetApprovedCommentsDynamicFunctionReturnTypeExtension.php
+++ b/src/GetApprovedCommentsDynamicFunctionReturnTypeExtension.php
@@ -1,0 +1,95 @@
+<?php
+
+/**
+ * Set return type of get_approved_comments() based on its passed arguments.
+ */
+
+declare(strict_types=1);
+
+namespace SzepeViktor\PHPStan\WordPress;
+
+use PhpParser\Node\Expr\FuncCall;
+use PHPStan\Analyser\Scope;
+use PHPStan\Reflection\FunctionReflection;
+use PHPStan\Type\ArrayType;
+use PHPStan\Type\Constant\ConstantStringType;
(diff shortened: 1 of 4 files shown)


src/GetApprovedCommentsDynamicFunctionReturnTypeExtension.php
```diff
@@ -0,0 +1,95 @@
...
+	 * - Return `int[]` if `$fields = 'ids'`.
+	 * - Return `int` if `$count = true`.
+	 *
+	 * @link https://developer.wordpress.org/reference/functions/get_approved_comments/#parameters
+	 */
+	public function getTypeFromFunctionCall( FunctionReflection $functionReflection, FuncCall $functionCall, Scope $scope ): ?Type {
+		$args = $functionCall->getArgs();
+
+		if ( \count( $args ) < 2 ) {
+			return self::defaultType();
```
szepeviktor, line 48: Isn't is `return null`?

szepeviktor: I will merge it anyway!

szepeviktor: https://github.com/szepeviktor/phpstan-wordpress/blob/6fbfbdfc822e70a669cff31d87545fad043f5a55/src/GetTermsDynamicFunctionReturnTypeExtension.php#L64-L66

szepeviktor:
Thank you.

PR #223, opened 2024-06-10
Fix #221

IanDelMar:
#221 was caused by the `VoidType` accepting `NullType` from PHPStan version 1.10.50 onwards (see [this commit](https://github.com/phpstan/phpstan-src/commit/40c8fb2266f040158fb811ac13b21a3e179c022a)).

This PR fixes #221 by using `isVoid()` to check for `VoidType`. `isVoid()` does not accept `NullType`.

Additionally, this PR adds a test for action callbacks that return `null`. Before this fix, the additional test was failing as well for the same reason.

Related: #222

PR #224, opened 2024-06-12
Allow returning null from a callback

### LINKED ISSUE #221: `__return_null()` in `add_filter()` causes error
`__return_null()` in `add_filter()` causes a `Filter callback return statement is missing.` error. See #220. Is this intended?

Note, there is no error with the lowest supported version of PHPStan.

om4csaba:
Fixes #221

Attempt to distinguish between returning void and null from a callback.

Unfortunately, some filters are expecting a null value. See an example in the WooCommerce repository here:

https://github.com/woocommerce/woocommerce/blob/7f657cca539af928a58b05716423971b737da91f/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L934-L938

### DIFF
--- a/src/HookCallbackRule.php
+++ b/src/HookCallbackRule.php
@@ -212,7 +212,7 @@ class HookCallbackRule implements \PHPStan\Rules\Rule
     {
         $returnType = $callbackAcceptor->getReturnType();
 
-        if ($returnType instanceof MixedType) {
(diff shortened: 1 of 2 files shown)


om4csaba:
Hi Viktor,

I missed that #223 already exists. Feel free to close this.

szepeviktor:
Thank you, Csaba.

PR #225, opened 2024-06-22
Sync with master

IanDelMar:


szepeviktor:
Thank you.
What are 2.x features?

IanDelMar:
The "feature" of version 2.x is that it does not come with extensions that are covered in https://github.com/php-stubs/wordpress-stubs, which means it requires fewer extensions and less maintenance.

szepeviktor:
What would be the right time to go v2?
https://wordpress.org/about/stats/

IanDelMar:
In terms of completeness, it will possibly take me a couple more weeks to have a deep dive into which currently provided extensions can be transferred to the stubs. I will let you know when I have finished that.

In terms of version support, it depends. But I think nothing is holding us back. I would argue that people using outdated versions of WordPress should simply not update their szepeviktor/phpstan-wordpress versions. At some point, you have to either accept that you are stuck on outdated versions or move on. :shrug:

PR #226, opened 2024-06-26
Add error identifiers

IanDelMar:
This PR introduces rule error identifiers for the following rules:
- `IsWpErrorRule`
- `HookDocsRule`
- `HookCallbackRule`

The identifiers are sourced from the PHPStan code base, except for `paramTag.count`. PHPStan does not have a corresponding error identifier for this; there is `arguments.count`, but this handles arguments, not tags. If anyone has better ideas on how to name the identifiers, I am happy to change them.

For `HookDocsRule` and `HookCallbackRule`, this PR replaces the try-catch blocks with checks for existing errors in an array holding the errors and removes the exceptions since they are no longer used. Keeping the exceptions would have required different exceptions for each error identifier to retrieve the correct identifier if the exception is thrown.

Changes not related to error identifiers:
- Introduces constants to give meaning to certain numbers.
- Reduces nesting level in `HookCallbackRule::getAcceptedArgs()`.
- Calls `HookCallbackRule::validateActionReturnType()` and `HookCallbackRule::validateFilterReturnType()` on the return type instead of on the callable acceptor.
- Renames `HookCallbackRule::isNever()` to `HookCallbackRule::isExplicitNever()` to be more explicit about the method's functionality.


szepeviktor:
Do we need to bump phpstan version?

IanDelMar:
I don't think so. `RuleErrorBuilder::message(...)->identifier(...)` has been around for a while now. See [this link](https://github.com/phpstan/phpstan-src/blob/6bdc25a944690839507a8ca96d33ec7c51660f3f/src/Rules/RuleErrorBuilder.php#L171-L177). The new interface [IdentifierRuleError](https://github.com/phpstan/phpstan-src/blob/466e6771a0899ac001dabc3984a49ef51c63d993/src/Rules/IdentifierRuleError.php) extends `RuleError` and simply adds a method to retrieve the identifier. Older versions of PHPStan don't know about this interface and do not use the `getIdentifier()` method. Newer versions of PHPStan complain about using the `RuleError` interface instead of the `IdentifierRuleError` interface.

It should be sufficient to ignore these errors.

szepeviktor:
Thank you, Ian.

PR #231, opened 2024-06-28
Fix/update extension for wp_parse_url

IanDelMar:
This update to `WpParseUrlFunctionDynamicReturnTypeExtension` from [PHPStan's extension for `parse_url`](https://github.com/phpstan/phpstan-src/blob/877ff0a620dc407182e71955fbde8b6e081a15ce/src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php) fixes an incorrect return type when the `$component` parameter is a non-constant integer. Additionally, the extension is now more specific about the port type.

### DIFF
--- a/src/WpParseUrlFunctionDynamicReturnTypeExtension.php
+++ b/src/WpParseUrlFunctionDynamicReturnTypeExtension.php
@@ -66,6 +66,8 @@ final class WpParseUrlFunctionDynamicReturnTypeExtension implements \PHPStan\Typ
 
         $this->cacheReturnTypes();
 
+        $componentType = new ConstantIntegerType(-1);
+
         if (count($functionCall->getArgs()) > 1) {
             $componentType = $scope->getType($functionCall->getArgs()[1]->value);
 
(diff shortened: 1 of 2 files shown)


src/WpParseUrlFunctionDynamicReturnTypeExtension.php
```diff
@@ -128,18 +145,18 @@ private function cacheReturnTypes(): void
...
         }
 
-        $stringType = new StringType();
-        $integerType = new IntegerType();
-        $falseType = new ConstantBooleanType(false);
-        $nullType = new NullType();
+        $string = new StringType();
+        $port = IntegerRangeType::fromInterval(0, 65535);
+        $false = new ConstantBooleanType(false);
+        $null = new NullType();
```
szepeviktor, line 151: Naming variable reserved words ...

szepeviktor:
GitHub is a bit down https://www.githubstatus.com/incidents/9vwllhs2w1kj

IanDelMar:
✅

PR #234, opened 2024-08-23
Add WPINC constant

mikemanger:
`WPINC` is not overridable but often used to include core files for functions such as `fetch_feed()`

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -11,6 +11,7 @@ define('WP_DEBUG_LOG', true);
 define('WP_DEBUG_DISPLAY', true);
 define('WP_PLUGIN_DIR', './');
 define('WPMU_PLUGIN_DIR', './');
+define('WPINC', 'wp-includes');
 define('EMPTY_TRASH_DAYS', 30 * 86400);
 define('SCRIPT_DEBUG', false);
 define('WP_LANG_DIR', './');


szepeviktor:
Thank you for your PR.

How do you start WordPress without including `wp-includes/feed.php`?
https://github.com/WordPress/WordPress/blob/83c20455880fae0f11ed823b8eb9d4a8640d0255/wp-settings.php#L231

mikemanger:
Oh you're right, the include isn't needed (thanks!). I think we just assumed it wasn't included because it is in the [example in the dev documentation](https://developer.wordpress.org/reference/functions/fetch_feed/#comment-681), I'll add a feedback note mentioning this.

There is a mention about adding `WPINC` [in the WordPress performance lab plugin](https://github.com/WordPress/performance/pull/1198#discussion_r1589801984) for their test suite but I'm not sure if it is needed as tests already get excluded from a lot of PHPStan things.

szepeviktor:
I know there is no core function to get the value of `WPINC` but I do not want to encourage the users of this package to use `WPINC`.

IanDelMar:
What about [includes_url()](https://developer.wordpress.org/reference/functions/includes_url/)?

mikemanger:
Lets close this as I was mistaken for our original needs. I agree it has limited use and encourages bad practices.

Issue #242, opened 2024-10-08
Prepare for PHPStan 2.0


swissspidy:
* https://github.com/phpstan/phpstan/discussions/11816
* https://github.com/phpstan/phpstan-src/blob/2.0.x/changelog-2.0.md
* https://github.com/phpstan/phpstan-src/blob/2.0.x/UPGRADING.md

Notable changes:

* Requires PHP 7.4+

szepeviktor:
> Requires PHP 7.4+

I will go for PHP 4.0 compatibility 🙃

IanDelMar:
I am wondering whether we should leave the master branch and go for PHPStan in the 2.x branch.

swissspidy:
Ah, was not aware of that branch. Yeah that could work.

PR #235, opened 2024-08-23
WIP: Move several extensions to WP stubs

IanDelMar:
### Description

This pull request syncs the 2.x branch with the master branch and removes extensions whose functionality has already been implemented or is about to be implemented by [php-stubs/wordpress-stubs](https://github.com/php-stubs/wordpress-stubs), or is no longer needed or is incorrect:

- `EchoKeyDynamicFunctionReturnTypeExtension`
- `GetApprovedCommentsDynamicFunctionReturnTypeExtension`
- `GetListTableDynamicFunctionReturnTypeExtension`
- `GetPostDynamicFunctionReturnTypeExtension`
- `GetPostsDynamicFunctionReturnTypeExtension`
- `GetSitesDynamicFunctionReturnTypeExtension`
- `GetTermsDynamicFunctionReturnTypeExtension`
- `WpDieDynamicFunctionReturnTypeExtension`
- `WpThemeMagicPropertiesClassReflectionExtension`

#### EchoKeyDynamicFunctionReturnTypeExtension
* Only `the_title_attribute()`, `wp_dropdown_languages()`, `wp_get_archives()`, `wp_list_bookmarks()`, `wp_list_categories()`, and `wp_list_pages()` have been transferred to php-stubs/wordpress-stubs.
* `wp_dropdown_categories()`, `wp_dropdown_pages()`, `wp_dropdown_users()`, and `wp_link_pages()` always return a string as documented, and hence are not included in any PR porting functionality to php-stubs/wordpress-stubs.
* The `$args` parameter of `get_search_form()`, `wp_list_authors()`, `wp_list_comments()`, `wp_list_users()`, `wp_login_form()`, and `wp_page_menu()` is filterable, and thus the value of `$args['echo']` is unknown. Therefore, these were excluded from the functionality ported to php-stubs/wordpress-stubs.

#### RedirectCanonicalDynamicFunctionReturnTypeExtension
Starting from PHPStan 1.10.49, void types, including void in unions, are transformed into null. This extension is no longer needed.

### Related

PRs implementing the functionality in [php-stubs/wordpress-stubs](https://github.com/php-stubs/wordpress-stubs)
- https://github.com/php-stubs/wordpress-stubs/pull/127
- https://github.com/php-stubs/wordpress-stubs/pull/195
- https://github.com/php-stubs/wordpress-stubs/pull/197
- https://github.com/php-stubs/wordpress-stubs/pull/198
- https://github.com/php-stubs/wordpress-stubs/pull/199
[...]

PR #243, opened 2024-10-09
PHPStan 2.0 support

swissspidy:
🤞 

- Bumps PHPStan requirement to 2.0+
- Bumps PHP requirement to 7.4+ accordingly
- Follows upgrade guide to pass new arguments where needed

There are currently 13 failing tests, all regarding `DynamicReturnTypeExtension`.

See #242


swissspidy:
> There are currently 13 failing tests, all regarding `DynamicReturnTypeExtension`.

Any volunteers?

swissspidy:
Many failures are related to `EchoKeyDynamicFunctionReturnTypeExtension` specifically. Seems like the `::getDefaultReturnType` method doesn't return the correct default return type here:

https://github.com/szepeviktor/phpstan-wordpress/blob/fdc31b2a37515acbe03069baf46489f88d0a4821/src/EchoKeyDynamicFunctionReturnTypeExtension.php#L181-L185

For example, for `wp_get_archives` it should be `string|void` as it can output both, but I only get `void`.

In the stubs the definition for that function is `@phpstan-return ($args is array{echo: false|0} ? string : void)`.

Maybe that's incorrect and it should be `@phpstan-return ($args is array{echo: false|0} ? string|void : void)` though? It's one of the functions that can always potentially return `void`, even if `echo=false`.

IanDelMar:
> Any volunteers?

I'll have a look at this.

szepeviktor:
So this should be switched to `2.x` branch.

IanDelMar:
The test for `_get_list_table()` fails due to the current version of the stubs. As of version 6.6.0, the stubs do not permit unknown types to be passed to `_get_list_table()`. An unknown type, or more specifically, any type other than `class-string`, will result in a PHPStan error concerning the passed argument.

swissspidy:
> So this should be switched to `2.x` branch.

Done (I hope)

@IanDelMar feel free to make/suggest any code changes

src/HookDocsRule.php
```diff
@@ -216,7 +216,7 @@ protected function validateSingleParamTag(string $paramName, ParamTag $paramTag,
             $this->currentScope->isDeclareStrictTypes()
         );
 
-        if (! $accepted) {
+        if ($accepted->result) {
```
swissspidy, line 219:
```suggestion
        if (! $accepted->result) {
```

I think. Oversight from rebasing?

src/AssertWpErrorTypeSpecifyingExtension.php
```diff
@@ -38,7 +38,12 @@ public function specifyTypes(MethodReflection $methodReflection, MethodCall $nod
     {
         $args = $node->getArgs();
 
-        return $this->typeSpecifier->create($args[0]->value, new ObjectType('WP_Error'), TypeSpecifierContext::createTruthy());
+        return $this->typeSpecifier->create(
+            $args[0]->value,
+            new ObjectType(\WP_Error::class),
```
swissspidy, line 43: Or is `ObjectType('WP_Error')` preferred?

szepeviktor:
This is so COMPLEX!
I hope everything has a **test**.

src/AssertWpErrorTypeSpecifyingExtension.php
szepeviktor: Less strings -> brighter future 🌻

IanDelMar:
@swissspidy You need to remove PHP 7.2 from travis.yml

szepeviktor:
We have only a few problems in CI.

szepeviktor (approved):

IanDelMar:
PR with fix for `wp_get_archives()` in https://github.com/php-stubs/wordpress-stubs/

IanDelMar:
> We have only a few problems in CI. Very few.

I guess that most or all of them would be resolved if we merge #235

szepeviktor:
> I guess that most or all of them would be resolved if we merge #235

Done 🍏

IanDelMar:
@swissspidy, could you rebase your PR so we can see what still needs to be resolved?

IanDelMar:
All errors, except for a few cases related to `wp_get_archives()` that have already been resolved in the stubs, arise from restricting argument types in the stubs in combination with conditional return types that modify the default return value. If the argument types are restricted in the stubs, then—knowing that PHPStan will handle this—the stubs do not need to account for the return type of an invalid argument.

For `_get_list_tables()`, see [this comment](https://github.com/szepeviktor/phpstan-wordpress/pull/243#issuecomment-2403482830). The other errors result from restricting the argument to arrays.

Testing with an invalid argument type, such as a query string, may lead to unexpected return types depending on the default return value of the conditional return type (i.e. the `else` branch that is reached if none of the `if` or `elseif` conditions are met). If PHPStan encounters `'echo=1&akey=avalue'`, it is considered a string, and the conditional return types `($args is array{echo: false|0} ? string : void)` and `($args is array{echo: true|1} ? string : void)` both evaluate to false (string ≠ array). The solution is to avoid testing with invalid argument types.

This issue does not occur with all functions with restricted arguments in the `echo` key extension because not all extensions have been ported to the stubs (i.e. not all have conditional return types). Some were incorrect (I submitted PRs to remove these extensions), and some have not been ported because `$args` is filterable, meaning we cannot determine the actual value or type of `$args`.

As there is no `echo` key extension left in v2.x, we only need to address this in the master branch. I suggest removing the tests that fail due to what I have just described.

szepeviktor:
We have done a lot work.
Ian: keyboard
Viktor: mouse

tests/HookDocsRuleTest.php
```diff
@@ -26,6 +26,7 @@ protected function getRule(): Rule
         return new HookDocsRule($fileTypeMapper, $ruleLevelHelper);
     }
 
+    // phpcs:ignore SlevomatCodingStandard.Functions.FunctionLength.FunctionLength
```
IanDelMar, line 29:
```suggestion
-    // phpcs:ignore SlevomatCodingStandard.Functions.FunctionLength.FunctionLength
```

tests/HookCallbackRuleTest.php
```diff
@@ -17,6 +17,7 @@ protected function getRule(): \PHPStan\Rules\Rule
         return new HookCallbackRule();
     }
 
+    // phpcs:ignore SlevomatCodingStandard.Functions.FunctionLength.FunctionLength
```
IanDelMar: This has been moved to the config file and can be removed here.

tests/HookDocsRuleTest.php
IanDelMar: This has been moved to the config file and can be removed here.

szepeviktor:
Are we ready in this PR?

swissspidy:
I think so. No idea what's wrong with that `Codacy Static Code Analysis` check though.

IanDelMar:
`stdout: ERROR: Referenced sniff "PSR12NeutronRuleset" does not exist` :open_mouth:

PR #247, opened 2024-10-10
Remove return type extension for wp_dropdown_pages

IanDelMar:
As documented, `wp_dropdown_pages()` always returns a string. See lines 1245-1249 in [wp-includes/post-template.php](https://github.com/WordPress/WordPress/blob/e25ffd14269ad6461aba63e3119ac5603502bf06/wp-includes/post-template.php#L1245-L1249).


szepeviktor:
My computer mouse suffers a serious injury!!

PR #254, opened 2024-11-11
Add type declarations to properties

IanDelMar:
v2 requires PHP 7.4, so we can now add type declarations to properties.


szepeviktor:
👍🏻

PR #258, opened 2024-11-12
Delete ApplyFiltersTestClass.php

IanDelMar:
It looks like the class `ApplyFiltersTestClass` is not used. Could someone confirm this?

A PHPUnit run with and without this file also produces the same test and assertion count (58) with no errors.


szepeviktor:
Is this file enough?
https://github.com/szepeviktor/phpstan-wordpress/blob/2.x/tests/data/apply_filters.php

PR #259, opened 2024-11-12
Include ApplyFiltersTestClass in tests

szepeviktor:
@IanDelMar This is a counter-PR 😃

What do you think?


IanDelMar:
We could do that, but I’m not sure it adds value to the extensive list of assertions in [tests/data/apply_filters.php](https://github.com/szepeviktor/phpstan-wordpress/blob/efbc730a469ef8a00a22e5acc09afe864d7e4162/tests/data/apply_filters.php).

IanDelMar:
@szepeviktor The class is also orphaned in v1.

szepeviktor:
> The class is also orphaned in v1.

Thank you again.
Fixed in 079942a

PR #261, opened 2024-11-14
Add support for named arguments in add_filter and add_action

### LINKED ISSUE #194: Named arguments break add_filter/add_action callback check
**Reproduce:**
```php
add_filter('filter', function($one, $two) {
    return 123;
}, accepted_args: 2);
```

**Expected:**
No errors

**Actual**
```
Callback expects 2 parameters, $accepted_args is set to 1.
```

IanDelMar:
Closes #194

> [!NOTE]  
> "WordPress does not support named parameters."
See [Core Handbook](https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/)

WordPress does not support named arguments, which means that some issues are not yet resolved (see ticket [PHP 8.0: improvements to allow for named parameters](https://core.trac.wordpress.org/ticket/59649) for details). However, `add_filter` and `add_action` are unaffected by these issues.

### DIFF
--- a/src/HookCallbackRule.php
+++ b/src/HookCallbackRule.php
@@ -14,6 +14,7 @@ use PhpParser\Node\Expr\FuncCall;
 use PHPStan\Analyser\Scope;
 use PHPStan\Reflection\ParametersAcceptor;
 use PHPStan\Reflection\ParametersAcceptorSelector;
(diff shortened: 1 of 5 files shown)


szepeviktor:
Yes 🫡 Sir!

PR #265, opened 2024-11-16
Fix and improve shortcode_atts extension

IanDelMar:
The current `shortcode_atts` function extension unconditionally adds string types to all constant `$pairs` array values. This behaviour is overly broad, as string type values should only be added for keys that are also present in `$atts`. Additionally, it does not account for cases where `$pairs` is a general array or `$atts` is an array shape. Consequently, this results in imprecise and partially inaccurate type inference in such scenarios.

Changes introduced by this PR:
- Support for non-constant `$pairs`,
- Accurate handling of empty `$atts`, and
- Improved merging logic.


szepeviktor:
Thank you.

PR #266, opened 2024-11-16
Remove testing wp_parse_url without argument

IanDelMar:
`$url` is required (see [function reference](https://developer.wordpress.org/reference/functions/wp_parse_url/)).

### DIFF
--- a/tests/data/wp_parse_url.php
+++ b/tests/data/wp_parse_url.php
@@ -17,12 +17,6 @@ $integer = doFoo();
 /** @var string $string */
 $string = doFoo();
 
-/**
- * wp_parse_url()
- */
-$value = wp_parse_url();
-assertType('mixed', $value);
-
 $value = wp_parse_url('http://abc.def');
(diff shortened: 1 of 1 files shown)


szepeviktor:
Thank you for spotting this.

- We should raise an error https://github.com/szepeviktor/phpstan-wordpress/blob/079942af95124e7e3bf95ab6ef2e006eba02e4e1/src/WpParseUrlFunctionDynamicReturnTypeExtension.php#L63-L65
- Test that error

Do you agree?

szepeviktor:
Halfdone.

We already have an error.

```
Function wp_parse_url invoked with 0 parameters, 1-2 required.
```

Why does this assertion not fail??
Because this is not static analysis but unit testing?

IanDelMar:
> Because this is not static analysis but unit testing?

Yes. See the [TypeInferenceTestCase](https://github.com/phpstan/phpstan-src/blob/4f33a8ea09c0fbacb9705b12a2327beba7f96463/src/Testing/TypeInferenceTestCase.php#L113-L130). Only the return type is checked.

Issue #197, opened 2023-08-07
Stop non-existent method/function in remove_action()


szepeviktor:
```php
remove_action('wp_enqueue_scripts', [$theme, 'wp_common_block_scripts_and_styles']);
```

This triggered no error on level 5.
How could that be?

johnbillion:
This is expected behaviour. The `$callback` parameter doesn't have to be valid.

```php
 * @param callable|string|array $callback  The name of the function which should be removed.
 *                                         This function can be called unconditionally to speculatively remove
 *                                         a callback that may or may not exist.
```

szepeviktor:
Thank you, John.

(For extraterrestrial aliens 👽) could you add something like `HookCallbackRule` that checks for callback existence in hooking function?

johnbillion:
I'd rather not 🤣

PR #268, opened 2024-11-17
Introduce conditionalTags

IanDelMar:
Allows rules to be disabled. Rules are enabled by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,40 @@ PHPStan interprets `$title` as `string`.
 
 For best results, ensure the first `@param` tag in these docblocks is accurate.
 
+## Disabling Rules
+
+You can disable rules using configuration parameters:
+
+```neon
+parameters:
+    phpstanWP:
+        checkHookCallback: false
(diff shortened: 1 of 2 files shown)


szepeviktor:
Thank you for this PR.

Do we work on this package because we want users partially use it?

IanDelMar:
I was considering more 'controversial' rules that we might want to add (e.g., your suggestion in #197 or possibly enforcing DocBlocks for hooks). If we apply this approach to some rules, why not apply it to all rules?

szepeviktor:
> more 'controversial' rules

Using static analysis for "'code'" (scripts) written for WordPress is 100% controversial.

szepeviktor:
What we three see okay for this package that must be used by all.

szepeviktor:
> Using static analysis for "'code'" (scripts) written for WordPress is 100% controversial.

@IanDelMar Have you seen what others come up with before this package? Tragedy!

PR #276, opened 2024-12-13
Add rule to suggest replacing WP constants with function calls

IanDelMar:
Adds a rule that checks for the use of certain WordPress constants. If such constants are found, an error will be triggered, suggesting replacing the constant with a function call.

I also suggest removing the following lines:  
https://github.com/szepeviktor/phpstan-wordpress/blob/6686dc2412cdf544556d428e67aa8f0b645164ca/bootstrap.php#L69C1-L72C49  
as `STYLESHEETPATH` and `TEMPLATEPATH` were deprecated in WordPress v6.4.0. Additionally, these constants may need to be added to `dynamicConstantNames` to avoid always-true and always-false errors.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -32,6 +32,7 @@ services:
 rules:
     - SzepeViktor\PHPStan\WordPress\HookCallbackRule
     - SzepeViktor\PHPStan\WordPress\HookDocsRule
+    - SzepeViktor\PHPStan\WordPress\WpConstantFetchRule
 parameters:
     bootstrapFiles:
         - ../../php-stubs/wordpress-stubs/wordpress-stubs.php
--- a/src/WpConstantFetchRule.php
+++ b/src/WpConstantFetchRule.php
@@ -1,0 +1,91 @@
+<?php
+
+/**
+ * Custom rule to discourage using WordPress constants fetchable via functions.
+ */
+
+declare(strict_types=1);
+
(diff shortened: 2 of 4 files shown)


szepeviktor:
Are we in the space age? 🙃

IanDelMar:
> Additionally, these constants may need to be added to `dynamicConstantNames` to avoid always-true and always-false errors.

If this is the case, it would apply to all other constants that are likely to have values different from those defined in `bootstrap.php`.

szepeviktor:
> I also suggest removing the following lines:
> https://github.com/szepeviktor/phpstan-wordpress/blob/6686dc2412cdf544556d428e67aa8f0b645164ca/bootstrap.php#L69C1-L72C49
> as `STYLESHEETPATH` and `TEMPLATEPATH` were deprecated in WordPress v6.4.0. Additionally, these constants may need to be added to `dynamicConstantNames` to avoid always-true and always-false errors.

https://github.com/szepeviktor/phpstan-wordpress/blob/6686dc2412cdf544556d428e67aa8f0b645164ca/composer.json#L15

@westonruter it was your PR.
What do you think about removing them from 2.x?

westonruter:
> What do you think about removing them from 2.x?

Make it so.

szepeviktor:
Going in ...

PR #279, opened 2025-02-12
Update examples

IanDelMar:
Removes the following files as the examples are outdated:  
- `fix-callable.php`: PHPStan recognises `$callable = [ $instance, $method ];` as a callable  
- `fix-parse-url.php`: PHPStan knows the [return type of `parse_url`](https://github.com/phpstan/phpstan-src/blob/2.1.x/src/Type/Php/ParseUrlFunctionDynamicReturnTypeExtension.php)  
- `fix-wpdb.php`: the stubs file adds `global $wpdb;`  

Updates the following files:  
- `phpstan-full.neon`: WordPress versions lower than 6.6 are not supported  
- `impure-have-functions.stub`: the stubs file informs PHPStan that `have_posts` and `WP_Query::have_posts` are impure  
- `composer.json`: bump the required PHP version to versions supported by the current version of `phpstan-wordpress`; bump the version of `phpstan-wordpress` to prevent incorrect requirements via copy and paste

### DIFF
--- a/examples/phpstan-full.neon
+++ b/examples/phpstan-full.neon
@@ -26,15 +26,6 @@ parameters:
     ignoreErrors:
         # Uses func_get_args()
         - '#^Function apply_filters(_ref_array)? invoked with [34567] parameters, 2 required\.$#'
-        # Fixed in WordPress 5.3
-        #- '#^Function do_action(_ref_array)? invoked with [3456] parameters, 1-2 required\.$#'
-        #- '#^Function current_user_can invoked with 2 parameters, 1 required\.$#'
-        #- '#^Function add_query_arg invoked with [123] parameters?, 0 required\.$#'
-        #- '#^Function wp_sprintf invoked with [23456] parameters, 1 required\.$#'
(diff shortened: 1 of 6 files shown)


examples/phpstan-full.neon
```diff
@@ -26,15 +26,6 @@ parameters:
     ignoreErrors:
         # Uses func_get_args()
         - '#^Function apply_filters(_ref_array)? invoked with [34567] parameters, 2 required\.$#'
```
szepeviktor, line 28: This could also go.

szepeviktor:
Touching relics?? :)

examples/phpstan-full.neon
IanDelMar: I'm not sure about `apply_filters_ref_array`, that's why I did not remove it. And yes, `apply_filters` could be removed.

szepeviktor:
Thank you.

IanDelMar:
You're very welcome.

PR #284, opened 2025-06-21
Remove `WP_CLI::error` from earlyTerminatingMethodCalls config

### LINKED ISSUE #213: WP_CLI::error() doesn't always terminate
The `WP_CLI::error()` method [is marked as an early terminating method](https://github.com/szepeviktor/phpstan-wordpress/blob/f7ff091331bc00c5688fe4ce0c4d51d06fa61553/extension.neon#L155), but it doesn't terminate if the `$exit` parameter is set to `false`.

Its entry in the `earlyTerminatingMethodCalls` config should be removed and replaced with something else that's conditional, but I'm not sure what.

lipemat:
`WP_CLI::error` supports an `$exit` parameter to disable terminating.

Fixes: #213

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -55,5 +55,4 @@ parameters:
         WP_Ajax_Response:
             - send
         WP_CLI:
-            - error
             - halt


szepeviktor:
Thank you!

szepeviktor:
@johnbillion Will it be a nicer day with this change?

johnbillion:
I don't think this is right approach because the method terminates by default. It needs to be a conditional rule based on the value of the `$exit` parameter.

szepeviktor:
@lipemat Could you make John's wish come true?
I think it needs a PHPStan extension.

lipemat:
Ideally we would remove both `WP_CLI:` from `earlyTerminatingMethodCalls` in this extension in favor of having the proper `@return` types in the WP_CLI stubs. 

We have an effort to add proper types happening [over here](https://github.com/wp-cli/wp-cli/pull/6096). Until the `earlyTerminatingMethodCalls` are removed from the `phpstan-wordpress` extension, the new `@phpstan-return` type will not work.

lipemat:
The proper types made it into the [main branch](https://github.com/wp-cli/wp-cli/blob/cbb6a51cf182aec2aac386e0dc40ee77ac25dc2e/php/class-wp-cli.php#L915) on wp-cli.

szepeviktor:
Very nice.
I merge this one.

Let's hope the Internet won't blow up!

PR #287, opened 2025-07-07
Remove `SAVEQUERIES` from the `dynamicConstantNames` list

IanDelMar:
This PR removes the `SAVEQUERIES` constant from the `dynamicConstantNames` list in `extension.neon`, as it is not defined by WordPress (see this [GitHub search](https://github.com/search?q=repo%3AWordPress%2Fwordpress+SAVEQUERIES&type=code)).

Including `SAVEQUERIES` in the `dynamicConstantNames` list prevents PHPStan from reporting the following code:
```php
if (SAVEQUERIES) {
    // do something
}
```
If `SAVEQUERIES` is not defined, this will trigger a fatal error in PHP 8+.

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -44,7 +44,6 @@ parameters:
         - EMPTY_TRASH_DAYS
         - WP_CLI
         - COOKIE_DOMAIN
-        - SAVEQUERIES
         - SCRIPT_DEBUG
     earlyTerminatingFunctionCalls:
         - wp_send_json


szepeviktor:
#172 What do you think @johnbillion ?
🔎 https://github.com/search?q=owner%3Ajohnbillion%20SAVEQUERIES&type=code

johnbillion (approved):
Looks good. I think that means that `WP_CLI` should be removed too though.

johnbillion:
Oh Ian beat me to it with #288.

szepeviktor:
I would leave it as is.

PR #290, opened 2025-07-07
Remove `STYLESHEETPATH` and `TEMPLATEPATH` from bootstrap.php

IanDelMar:
`STYLESHEETPATH` and `TEMPLATEPATH` have been deprecated in WP 6.4.0 and are handled via the `WpConstantFetchRule` rule.


szepeviktor:
Thank you.

PR #294, opened 2025-09-02
Add return type extension for stripslashes_from_strings_only()

IanDelMar:
See https://github.com/php-stubs/wordpress-stubs/issues/333


szepeviktor:
Widely used: https://grep.app/search?q=stripslashes_from_strings_only 🙃

PR #296, opened 2025-09-02
Add return type extension for wp_slash()

IanDelMar:
See https://github.com/php-stubs/wordpress-stubs/issues/333


szepeviktor:
In my dreams (DreamPress) functions are one-purpose and straight-forward: one type in, one type out. That's all.

PR #298, opened 2025-09-09
Add return type extension for normalize_whitespace

IanDelMar:
Adds return type extension for [`normalize_whitespace()`](https://developer.wordpress.org/reference/functions/normalize_whitespace/).

For constant strings:
- trims strings
- keeps empty and falsy strings
- generalises all other constant strings

For non-constant strings:
- converts `non-falsy-string` to `non-empty-string` (`' 0'` is no longer non-falsy after `trim()`)
- converts `non-empty-string` to `string` (`' '` is no longer non-empty after `trim()`)
- preserves all other `AccessoryType`s

### DIFF
--- a/extension.neon
+++ b/extension.neon
@@ -10,6 +10,10 @@ services:
         tags:
             - phpstan.broker.dynamicFunctionReturnTypeExtension
     -
+        class: SzepeViktor\PHPStan\WordPress\NormalizeWhitespaceDynamicFunctionReturnTypeExtension
+        tags:
+            - phpstan.broker.dynamicFunctionReturnTypeExtension
+    -
         class: SzepeViktor\PHPStan\WordPress\ShortcodeAttsDynamicFunctionReturnTypeExtension
         tags:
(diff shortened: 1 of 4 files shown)


IanDelMar:
The error is related to `uppercase-string`, introduced in PHPStan [1.12.12](https://github.com/phpstan/phpstan/discussions/12144). It appears some issues around it were only resolved in 2.0.3. Should we remove the `' FOO '` assertion, or bump the minimum required PHPStan version to 2.0.3? I think we can safely just remove the assertion.

IanDelMar:
Perhaps `uppercase-string` was introduced in v1 and not in v2 initially. From the release description of version [2.0.3](https://github.com/phpstan/phpstan/releases/tag/2.0.3):
> This release includes improvements and fixes from [PHPStan 1.12.12](https://github.com/phpstan/phpstan/releases/tag/1.12.12).

Still, we can just remove the assertion. The type is not actively used, but is part of the return types composed from the `AccessoryType`s by PHPStan.

szepeviktor:
🫡 abnormalize_whitespace

PR #299, opened 2025-09-11
Add slashit functions extension & fix stripslashes extension

IanDelMar:
Adds a dynamic function return type extension for handling:

1. [`untrailingslashit()`](https://developer.wordpress.org/reference/functions/untrailingslashit/) — removes trailing forward slashes and backslashes,
2. [`trailingslashit()`](https://developer.wordpress.org/reference/functions/trailingslashit/) — adds a trailing slash after removing any trailing slashes,
3. [`backslashit()`](https://developer.wordpress.org/reference/functions/backslashit/) — adds backslashes before letters and before a number at the start of a string.

For (1) and (2), the extension uses the types PHPStan provides for `rtrim()` and string concatenation.
For (3), the extension traverses types, generalising constant strings to strings with accessory types, and replaces `numeric-string` (backslashes will be added) and `non-empty-string` (if it were `'0'`, backslashes would be added) with `non-falsy-string`.

Fixes the `StripslashesFromStringsOnlyDynamicFunctionReturnTypeExtension` by letting PHPStan handle the type for non-constant string arguments (e.g. a non-falsy string might be `'\\'`, which becomes `''`). Currently, PHPStan returns `string` for calls to `stripslashes()`. If PHPStan ever decides that `stripslashes()` should be, or is worth being, handled more granularly, this extension will align with that.


szepeviktor:
It fails with `Locking phpstan/phpstan (2.0.0)`
How high should I raise it?

https://github.com/szepeviktor/phpstan-wordpress/blob/49b833d466e0acca18e96fc15ce9868f44d8de72/composer.json#L16

IanDelMar:
Since 2.1.18, there are no errors; the "errors" (nothing is incorrect--prior to 2.1.18 the types were simply wider) occurred because PHPStan improved its inference of `rtrim()`’s return type. It’s unfortunate that we have to bump the version to overcome this.

szepeviktor:
Done.

PR #301, opened 2025-09-13
Gather assert types conditional on installed PHPStan version

IanDelMar:
Types may depend on the PHPStan version used--especially when a dynamic return type extension relies on types inferred by PHPStan.

Example from `SlashitFunctionsDynamicFunctionReturnTypeExtension`, where the type depends on PHPStan’s handling of `rtrim()`:
https://github.com/szepeviktor/phpstan-wordpress/blob/1cdb6d5a1951904daf8d54df3f40e2fc5df45bce/src/SlashitFunctionsDynamicFunctionReturnTypeExtension.php#L62

Among other cases, this can cause tests to fail in the [lowest-deps](https://github.com/szepeviktor/phpstan-wordpress/blob/1cdb6d5a1951904daf8d54df3f40e2fc5df45bce/.travis.yml#L31-L38) build. To avoid bumping the required PHPStan version, this PR introduces the `InstalledPhpStanVersion` class and uses it in `DynamicReturnTypeExtensionTest` to avoid gathering types for versions that produce different results than the latest.

The types returned in the lowest supported version should still be checked to rule out incorrect cases. This has been done for `data/slashit-functions.php`. We could also add, e.g., `data/slashit-functions-bc.php` with assertions that pass with the lowest PHPStan version. This has not been done yet.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,10 +13,11 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "php-stubs/wordpress-stubs": "^6.6.2",
-        "phpstan/phpstan": "^2.1.18"
+        "phpstan/phpstan": "^2.0"
     },
     "require-dev": {
         "composer/composer": "^2.1.14",
+        "composer/semver": "^3.4",
         "dealerdirect/phpcodesniffer-composer-installer": "^1.0",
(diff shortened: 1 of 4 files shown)


szepeviktor:
That is _my Ian_!
Supporting 28 years old software!

PR #306, opened 2026-02-23
Add WP_START_TIMESTAMP constant

dannyvankooten:
`WP_START_TIMESTAMP` is a default WordPress constant defined in `wp-includes/default-constants.php` which was missing in this stub. This PR adds it so that phpstan can be used for code relying on that constant without having to exclude the rule.


szepeviktor:
Hello!

Thank you for this PR.
Please don't use internal stuff in WP. Timing has nice user functions.

